PR #207, opened 2015-01-20
Fix tax Portion money amount is null. Close #203

### LINKED ISSUE #203: Tax Portion money amount is null.
At least in the example I'm working with, it's always returning null, even though I'm sure the value is set.

```cart.getTaxedPrice().ifPresent(tp -> tp.getTaxPortions().stream().forEach(tt -> System.out.println(tt.getRate() + " " + tt.getMoney())));```

Returns `0.19 null`

schleichardt:


lauraluiz:
Nice, thanks!

PR #209, opened 2015-01-20
Create catchable exception when requested offset is greater than 100000,...

### LINKED ISSUE #198: Create catchable exception when requested offset is greater than 100000
When you request an offset which is greater than 100000 you receive the following error:

io.sphere.sdk.client.SphereBackendException: Malformed parameter: offset: The offset parameter must be in the range of [0..100000]

This issue might rarely occur in real-life, but it would be nice to be able to catch it without parsing the response-body. Maybe the solution could be the same as in 
https://github.com/sphereio/sphere-jvm-sdk/issues/178

schleichardt:
... close #198


hajoeichler:
:+1:

PR #238, opened 2015-01-26
Client instantiation

schleichardt:


lauraluiz:
Good. Merge and release :D

hajoeichler:
@schleichardt and @lauraluiz we should add a real world example where the credentials are either env variables or read from a config file (`.properties`?).

hajoeichler:
Sorry, wrong button

schleichardt:
Yep, the underlying task is not done by merging the pull request:

https://github.com/sphereio/sphere-jvm-sdk/issues/177

lauraluiz:
Yes, that would be good. Kind of the typical ways to initialize the client in a practical way, a whole section for that. But this leads us to a new matter, I think we should find a way to organize all the documentation that we have in a comprehensive way. Sometimes I find it difficult to find documentation that I know exists, and it's a shame, since we have many and good.

PR #251, opened 2015-02-02
Product update actions

schleichardt:
@lauraluiz please review.

Release notes are in http://sphereio.github.io/sphere-jvm-sdk/javadoc/product-update-actions/io/sphere/sdk/meta/ReleaseNotes.html

### DIFF
--- a/common/src/main/java/io/sphere/sdk/models/LocalizedStrings.java
+++ b/common/src/main/java/io/sphere/sdk/models/LocalizedStrings.java
@@ -22,7 +22,7 @@ import static java.util.stream.Collectors.joining;
  * Note that even if your project only uses one language some attributes (name and description for example) will be
  * always be LocalizedStrings.
  */
-public class LocalizedStrings {
+public class LocalizedStrings extends Base {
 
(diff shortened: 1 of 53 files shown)


common/src/main/java/io/sphere/sdk/models/LocalizedStrings.java
```diff
@@ -137,24 +137,10 @@ public String toString() {
...
-        if (o == null || getClass() != o.getClass()) return false;
-        LocalizedStrings localized = (LocalizedStrings) o;
-        return this.translations.equals(localized.translations);
-    }
-
-    @Override
-    public int hashCode() {
-        return translations.hashCode();
-    }
-
```
lauraluiz: Why are the equality methods not good here?

schleichardt: The class was mostly from the Play SDK, in the JVM SDK we just extend Base and do not reimplement equals, hashCode and toString.

lauraluiz: Oops true, I just forgot we had `Base` (as per usual) :D

lauraluiz:
Looks good!

PR #280, opened 2015-02-20
Feature/state and item state queries

abrauner:
I hope there are no more mistakes in it. If so I will update the PR. I added some tests according to the channels to create and delete a state. Would be great to get this into the SDK. 

**Note** 
there is a test in CartUpdateCommandTest set to @Ignore because here the ItemState has to be created first to get also a reference. 

Cheers


abrauner:
Travis complains about missing credentials when I'm right. Is this related to my build? Or a general problem?? Maybe I have to configure something in my Project?

schleichardt:
You work on a different repository, so for security reasons our encrypted credentials cannot be used by you. If the tests are green on your machine, everything is fine, don't worry about travis.

schleichardt:
@abrauner  Thank you very much! Do you want into the hall of fame? https://github.com/sphereio/sphere-jvm-sdk/blob/master/project/Release.scala#L75

abrauner:
Don't know if this is applicable for externals, but if so why not ;-)

schleichardt:
@hajoeichler Is it possible to put him into our hall of fame?

hajoeichler:
We should add a thx notice to the release notes and if @abrauner is further contributing, we will add him to the list of devs ... motivation is everything ;)

PR #292, opened 2015-03-10
Order import

schleichardt:


models/src/it/java/io/sphere/sdk/orders/commands/OrderImportCommandTest.java
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    @Test
+    public void getCompletedAt() throws Exception {
+        final Instant completedAt = Instant.now().minusSeconds(5555);
+        testOrderAspect(builder -> builder.completedAt(completedAt),
+                order -> assertThat(order.getCompletedAt()).isPresentAs(completedAt));
+    }
+
+    private void testOrderAspect(final Consumer<ImportOrderBuilder> orderBuilderConsumer, final Consumer<Order> orderConsumer) {
```
lauraluiz: Just a naming question: what do you mean with `Aspect`?

schleichardt: Yes, the naming is weird. I tested kind of every ImportAttribute field in an extra test method, since there were all the time issues and I wanted to work in small steps.
How could I improve it?

PR #297, opened 2015-03-11
Nested attribues

OlegIlyenko:
@schleichardt can you please review?


models/src/main/java/io/sphere/sdk/attributes/AttributeContainerImpl.java
```diff
@@ -0,0 +1,50 @@
...
+            } catch (final JsonException e) {
+                throw transformError(e, attributeName, mapper);
+            }
+        });
+    }
+
+    protected JsonException transformError(JsonException e, String attributeName, AttributeMapper<?> mapper) {
+//        throw new JsonException(format("TODO: %s does not contain an attribute '%s' which can be mapped with %s.", objectWithAttributes, attributeName, mapper), cause);
+        // TODO: implement proper error handling
+        return e;
```
OlegIlyenko: Sorry, forgot this one. I will add proper error handling here now.

OlegIlyenko: Fixed it here: https://github.com/sphereio/sphere-jvm-sdk/commit/e7390c607dcd4af39c0db49b389ce1d6c9497c49

models/src/it/java/io/sphere/sdk/producttypes/NestedAttributeIntegrationTest.java
```diff
@@ -0,0 +1,191 @@
...
+        static AttributeGetterSetter<Product, Set<AttributeContainer>> nutrientInformation = AttributeAccess.ofNestedSet().ofName("nutrientInformation");
+    }
+
+    interface Banana {
+        static AttributeGetterSetter<Product, String> bananaColor = AttributeAccess.ofText().ofName("bananaColor");
+        static AttributeGetterSetter<Product, Set<AttributeContainer>> nutrients = AttributeAccess.ofNestedSet().ofName("nutrients");
+    }
+
+    @Test
+    public void itShouldBePossibleToDeeplyNestNestedTypesInsideOfSetTypes() throws Exception {
```
schleichardt: This method is too long to be clean code. Extract methods, please.

models/src/it/java/io/sphere/sdk/producttypes/ProductTypeIntegrationTest.java
```diff
@@ -75,6 +76,35 @@ public void booleanAttribute() throws Exception {
     }
 
     @Test
+    public void nestedAttribute() throws Exception {
+        TestCleaner.withCleaner(cleaner -> {
```
schleichardt: I think it is not a good idea to introduce different testing styles, the TestCleaner is in competition with the different test fixture classes and they are not compatible with each other.

The Fixture classes are well, because they enable cleanup before the use, thats cool in case of crashes and also they hide how to delete data. With the TestCleaner you duplicate how to delete something.

 I would accept this pull request with the TestCleaner but remove it myself in master. Further pull requests with a TestCleaner will not be accepted. I don't different architectures for the same thing in one product. It should all look alike and use few concepts. If we use a lot of concepts, it is required for contributors to learn each concept before starting and I actually want to make it easy to contribute.

models/src/it/java/io/sphere/sdk/utils/CleanupFunction.java
```diff
@@ -0,0 +1,6 @@
+package io.sphere.sdk.utils;
+
+@FunctionalInterface
+public interface CleanupFunction<T> {
```
schleichardt: You don't need to declare a functional interface, you could use http://docs.oracle.com/javase/8/docs/api/java/util/function/Consumer.html .

models/src/main/java/io/sphere/sdk/attributes/AttributeContainerImpl.java
```diff
@@ -0,0 +1,50 @@
...
+import io.sphere.sdk.models.Base;
+import io.sphere.sdk.products.Product;
+
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+
+import static java.lang.String.format;
+
+public class AttributeContainerImpl extends Base implements AttributeContainer {
```
schleichardt: This class should not (yet) be public. You can move it to the package and make it package scope.

models/src/main/java/io/sphere/sdk/attributes/NestedType.java
```diff
@@ -0,0 +1,28 @@
...
+    private NestedType(final Reference<ProductType> typeReference) {
+        this.typeReference = typeReference;
+    }
+
+    public Reference<ProductType> getTypeReference() {
+        return typeReference;
+    }
+
+    @JsonCreator
+    public static NestedType of(final Reference<ProductType> typeReference) {
```
schleichardt: Use Referenceable here, please.

models/src/main/java/io/sphere/sdk/attributes/AttributeContainer.java
```diff
@@ -0,0 +1,36 @@
...
+     *
+     * @throws io.sphere.sdk.json.JsonException if the type of attribute cannot be parsed
+     *
+     * @param accessor declaration of the name and type of the attribute
+     * @param <T> the underlying type of the attribute
+     * @return the value of the attribute, or Optional.empty if absent
+     */
+    <T> Optional<T> getAttribute(final AttributeGetter<Product, T> accessor);
+
+    default boolean hasAttribute(String attributeName) {
```
schleichardt: If you like, you can create an hasAttribute(AttributeGetter) or create a ticket for it and assign it to me.

schleichardt:
@OlegIlyenko could you please add some Javadoc how to use nested attributes? It is a very complex topic so it is really necessary to document it well.

models/src/main/java/io/sphere/sdk/attributes/NestedType.java
lauraluiz: The one with Referenceable is below, this one has to be therefore removed, as it is unnecessary.

OlegIlyenko: Unfortunately, Jackson does not understand `Referenceable` so I still
needed to keep other method around.

schleichardt: No, like in the rest of the SDK @JsonCreator
 should be on the constructor.

schleichardt:
:relieved:

PR #301, opened 2015-03-13
Add warning in case of deprecation notice header

agourlay:
This PR propagates in the SDK the latest changes in the backend concerning deprecation notices.

From now on Sphere.io API will communicate deprecation notices to clients using the custom Http header ```X-DEPRECATION-NOTICE```.
In order to log the appearance of this header in the response I had to : 
- add the headers information in the ```HttpResponse``` object.
- migrate the current ```HttpHeaders``` class to use a Multimap instead of a Map.

Therefore the diff is much bigger than it should have been, I hope I haven't messed up too many things :bowtie:


agourlay:
:+1:

PR #308, opened 2015-03-17
Add parent to category query model. Fix #304.

### LINKED ISSUE #304: CategoryQuery: No predicate for field parent
Hi, 
when trying to use the build-in dsl for creating predicates, the predicate for the field parent is missing.

If works when I try it for id:
```java
final Predicate<Category> idIsNotPresent = CategoryQuery.model().id().isNotPresent();
```

But when I try to use it with the parent field, there is no option to create a predicate so I have to figure out the predicate by myself: 
```java
final SimplePredicate parentIsNotPresent = new SimplePredicate("parent is not defined");
```

I think it would be better if the dsl could be used consistently for all query-able fields.

schleichardt:


schleichardt:
@lauraluiz please review

PR #331, opened 2015-03-25
Fix URL encoding

schleichardt:
@lauraluiz please review.


lauraluiz:
Your code is good. I just also added a test for filters, which forced me to refactor the IT a bit.
Please review and merge it.

PR #335, opened 2015-03-29
Implementation of StateBuilder for Unit-Tests and UpdateActions for State

abrauner:
Please have a look at the StateBuilder which might be useful for Unit-Tests and the missing UpdateActions<State>. Not sure if SetTransitions is correct according the Array and List<State>.

### DIFF
--- a/models/src/main/java/io/sphere/sdk/states/commands/updateactions/package-info.java
+++ b/models/src/main/java/io/sphere/sdk/states/commands/updateactions/package-info.java
@@ -1,0 +1,5 @@
+/**
+ * {@doc.gen summary categories}
+ */
+
+package io.sphere.sdk.states.commands.updateactions;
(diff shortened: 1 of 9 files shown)


schleichardt:
Thank you very much!
I will apply some changes in another pull request.

models/src/main/java/io/sphere/sdk/states/commands/updateactions/package-info.java
```diff
@@ -0,0 +1,5 @@
+/**
+ * {@doc.gen summary categories}
```
butenkor: should be states, right?

schleichardt: Yes, I fixed it in https://github.com/sphereio/sphere-jvm-sdk/pull/344/files#diff-3c8dac3fa3b8753175c3b556f474c098L2

PR #356, opened 2015-04-15
Implement mapValue method for LocalizedString, it provides map functiona...

schleichardt:
...lity. #333


lauraluiz:
Good, merge :+1:

PR #367, opened 2015-04-21
Reviews

njacintoct:


models/src/it/java/io/sphere/sdk/reviews/ReviewsFixtures.java
```diff
@@ -0,0 +1,32 @@
...
+import io.sphere.sdk.customers.Customer;
+import io.sphere.sdk.products.Product;
+import io.sphere.sdk.reviews.commands.ReviewCreateCommand;
+import io.sphere.sdk.test.SphereTestUtils;
+import java.util.Random;
+
+import java.util.function.Consumer;
+
+/**
+ * Created by njacinto on 20/04/15.
```
yanns: please avoid default generated javadoc.

models/src/test/java/io/sphere/sdk/reviews/ReviewDraftTest.java
```diff
@@ -0,0 +1,142 @@
...
+    public void testGetScore() throws Exception {
+        assertTrue(ReviewImplTest.isScoreInValidRange(DEFAULT_REVIEW.getScore().get(), ReviewImplTest.DEFAULT_SCORE));
+    }
+
+    @Test
+    public void testWithTitle() throws Exception {
+        ReviewDraft review = ReviewDraft.of(ReviewImplTest.DEFAULT_PRODUCT_ID, ReviewImplTest.DEFAULT_CUSTOMER_ID);
+        assertFalse(review.getTitle().isPresent());
+        ReviewDraft reviewWith = review.withTitle(ReviewImplTest.DEFAULT_TITLE);
+        assertFalse(review==reviewWith);
```
hajoeichler: What does the coding guideline says about spaces in this case?
Shouldn't we use `equal` to compare objects?

models/src/main/java/io/sphere/sdk/reviews/Review.java
```diff
@@ -0,0 +1,53 @@
...
+    public Optional<String> getText();
+
+    public Optional<Float> getScore();
+
+    public default Reference<Review> toReference() {
+        return Reference.of(typeId(), getId());
+    }
+
+    public static String typeId(){
+        return "review"; // why not a number?
```
hajoeichler: @njacintoct Can you bring some light to the question in the comment please ;)

models/src/main/java/io/sphere/sdk/reviews/queries/ReviewsEndpoint.java
```diff
@@ -0,0 +1,8 @@
+package io.sphere.sdk.reviews.queries;
```
hajoeichler: @schleichardt and @njacintoct This class looks like a 100% copy of the class in package `commands` - is this intended?

models/src/test/java/io/sphere/sdk/reviews/ReviewDraftTest.java
njacintoct: Hello,

It’s not comparing objects. It’s comparing references.

Could you please send me the code guidelines?

Regards,

Nuno Jacinto

> On 22 Apr 2015, at 09:20, Hajo <notifications@github.com> wrote:
> 
> In models/src/test/java/io/sphere/sdk/reviews/ReviewDraftTest.java <https://github.com/sphereio/sphere-jvm-sdk/pull/367#discussion_r28849579>:
> 
> > +    @Test
> > +    public void testGetText() throws Exception {
> > +        assertEquals(ReviewImplTest.DEFAULT_TEXT, DEFAULT_REVIEW.getText().get());
> > +    }
> > +
> > +    @Test
> > +    public void testGetScore() throws Exception {
> > +        assertTrue(ReviewImplTest.isScoreInValidRange(DEFAULT_REVIEW.getScore().get(), ReviewImplTest.DEFAULT_SCORE));
> > +    }
> > +
> > +    @Test
> > +    public void testWithTitle() throws Exception {
> > +        ReviewDraft review = ReviewDraft.of(ReviewImplTest.DEFAULT_PRODUCT_ID, ReviewImplTest.DEFAULT_CUSTOMER_ID);
> > +        assertFalse(review.getTitle().isPresent());
> > +        ReviewDraft reviewWith = review.withTitle(ReviewImplTest.DEFAULT_TITLE);
> > +        assertFalse(review==reviewWith);
> What does the coding guideline says about spaces in this case?
> Shouldn't we use equal to compare objects?
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/sphereio/sphere-jvm-sdk/pull/367/files#r28849579>.
> 


-- 


<http://berlin.apidays.io>

commercetools GmbH
SPHERE.IO <http://www.sphere.io>  *The E-Commerce Platform.*

Adams-Lehmann-Strasse 44, 80797 München

www.commercetools.com

Amtsgericht München, HRB 161496

Geschäftsführer: Denis Werner, Dirk Hörig, Dr. Robert Zores

models/src/main/java/io/sphere/sdk/reviews/queries/ReviewsEndpoint.java
njacintoct: I just copy the structure used on the channels

models/src/main/java/io/sphere/sdk/reviews/Review.java
njacintoct: sorry, it was for me and I forgot to remove it.

models/src/main/java/io/sphere/sdk/reviews/queries/ReviewsEndpoint.java
schleichardt: Yes, it is intended to be used that way. The endpoint classes decode the type references for the JSON deserialization and the path for the endpoint. Since I don't want them exposed to the public API, they need to be package scope in two packages.

schleichardt: It is duplication for information hiding.

models/src/main/java/io/sphere/sdk/reviews/Review.java
```diff
@@ -0,0 +1,53 @@
...
+    public Optional<String> getAuthorName();
+
+    public Optional<String> getTitle();
+
+    public Optional<String> getText();
+
+    public Optional<Float> getScore();
+
+    public default Reference<Review> toReference() {
+        return Reference.of(typeId(), getId());
```
schleichardt: please change to
```
public default Reference<Review> toReference() {
        return Reference.of(typeId(), getId(), this);
}
```
I will fix this in channels, there is it also missing.

models/src/main/java/io/sphere/sdk/reviews/ReviewImpl.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public String getProductId() {
+        return productId;
+    }
+
+    public String getCustomerId(){
+        return customerId;
+    }
+
+    @Override
```
lauraluiz: Can you put the @override in all other methods, that are actually being overridden as well? I know it's a silly request, but I feel then it's clearer and easier to change the interface.

njacintoct: not silly, I'm used to the editor warning me about it

models/src/test/java/io/sphere/sdk/reviews/ReviewDraftTest.java
```diff
@@ -0,0 +1,142 @@
...
+        assertFalse(review.getText().isPresent());
+        ReviewDraft reviewWith = review.withText(ReviewImplTest.DEFAULT_TEXT);
+        assertFalse(review==reviewWith);
+        assertEquals(ReviewImplTest.DEFAULT_TEXT, reviewWith.getText().get());
+    }
+
+    @Test
+    public void testWithScore() throws Exception {
+        ReviewDraft review = ReviewDraft.of(ReviewImplTest.DEFAULT_PRODUCT_ID, ReviewImplTest.DEFAULT_CUSTOMER_ID);
+        assertFalse(review.getScore().isPresent());
```
lauraluiz: I would skip this assertion since you are getting the value later, what do you think?

njacintoct: it's checking that the default value is null. 
It's relevant if this is the expected behaviour, otherwise it can be removed.

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
```diff
@@ -0,0 +1,392 @@
...
+            {"Fifth author","Fifth review","This is my fifth comment",0.7f}
+    };
+
+    @Test
+    public void testCreate() throws Exception {
+        withProduct(client(), product ->
+            withCustomer(client(), customer -> {
+                Instant start = Instant.now().minusSeconds(SERVER_TIME_TOLERANCE_SEC); // time tolerance
+                Review review = client().execute(ReviewCreateCommand.of(ReviewDraft.of(product.getId(), customer.getId())
+                        .withAuthorName((String) REVIEWS[0][0]).withTitle((String) REVIEWS[0][1])
```
schleichardt: I would prefer a data holder class in test scope instead of casting everywhere. You also could put the comment data and scores there.

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
```diff
@@ -0,0 +1,392 @@
...
+
+import java.time.Instant;
+import java.time.format.DateTimeFormatter;
+import java.util.Arrays;
+import java.util.Optional;
+
+import static io.sphere.sdk.customers.CustomerFixtures.withCustomer;
+import static io.sphere.sdk.products.ProductFixtures.withProduct;
+import static io.sphere.sdk.reviews.ReviewsFixtures.withReview;
+import static org.junit.Assert.*;
```
schleichardt: In every other test class we are using fest assertions. Is there a need for the exception here?

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
```diff
@@ -0,0 +1,392 @@
...
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Created by njacinto on 17/04/15.
+ */
```
schleichardt: I think we should not keep this comment.

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
```diff
@@ -0,0 +1,392 @@
...
+    private static final Object[][] REVIEWS = {
+            {"First author","First review","This is my first comment",0.3f},
+            {"Second author","Second review","This is my second comment",0.4f},
+            {"Third author","Third review","This is my third comment",0.5f},
+            {"Fourth author","Fourth review","This is my fourth comment",0.6f},
+            {"Fifth author","Fifth review","This is my fifth comment",0.7f}
+    };
+
+    @Test
+    public void testCreate() throws Exception {
```
schleichardt: In JUnit 4 it is not practice to start test methods with "test".

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
njacintoct: I can also create 2 arrays, one for the strings an another for the doubles (it's not float)

schleichardt: A class is easier to understand and argue about than two parallel arrays.

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
njacintoct: I'm not understanding what you mean by need of exception here

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
```diff
@@ -0,0 +1,392 @@
...
+                        .withText((String) REVIEWS[0][2]).withScore((Float) REVIEWS[0][3])));
+                Instant afterCreation = Instant.now().plusSeconds(SERVER_TIME_TOLERANCE_SEC); // time tolerance
+                assertEquals(product.getId(), review.getProductId());
+                assertEquals(customer.getId(), review.getCustomerId());
+                assertEquals((String) REVIEWS[0][0], review.getAuthorName().get());
+                assertEquals((String) REVIEWS[0][1], review.getTitle().get());
+                assertEquals((String) REVIEWS[0][2], review.getText().get());
+                assertEquals((Float) REVIEWS[0][3], review.getScore().get());
+                assertNotNull(review.getId());
+                assertEquals(1, review.getVersion());
```
schleichardt: You should not check the version, it is not guaranteed that the version number increases by one.

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
```diff
@@ -0,0 +1,392 @@
...
+                Instant timestamp = review.getCreatedAt();
+                if (!timestamp.isAfter(start) || !timestamp.isBefore(afterCreation)) {
```
schleichardt: for this check you could extract a method because there is a lot of repetition with the next block.

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
```diff
@@ -0,0 +1,392 @@
...
+                            DateTimeFormatter.ISO_INSTANT.format(timestamp),
+                            DateTimeFormatter.ISO_INSTANT.format(start),
+                            DateTimeFormatter.ISO_INSTANT.format(afterCreation)));
+                }
+            })
+        );
+
+    }
+
+    @Test(expected = io.sphere.sdk.client.ErrorResponseException.class)
```
schleichardt: Why not use an import?

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
```diff
@@ -0,0 +1,392 @@
...
+    public void testCreate_OneReviewPerCustomerProductConstraint() throws Exception {
+        withProduct(client(), product ->
+                withCustomer(client(), customer -> {
+                    for (int i = 0; i < 2; i++) {
```
schleichardt: 2 is a magic number here.

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
```diff
@@ -0,0 +1,392 @@
...
+                            DateTimeFormatter.ISO_INSTANT.format(start),
+                            DateTimeFormatter.ISO_INSTANT.format(afterCreation)));
+                }
+            })
+        );
+
+    }
+
+    @Test(expected = io.sphere.sdk.client.ErrorResponseException.class)
+    public void testCreate_OneReviewPerCustomerProductConstraint() throws Exception {
```
schleichardt: This test is not SDK related, you don't need to test the backend until you really want to define the client behavior.

For example if a special error response code is present you could test this. May ask me in person after I am done with the review.

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
```diff
@@ -0,0 +1,392 @@
...
+                    for (int i = 0; i < 2; i++) {
+                        Review review = client().execute(new ReviewCreateCommand(ReviewDraft.of(product.getId(), customer.getId())
+                                .withAuthorName((String) REVIEWS[i][0]).withTitle((String) REVIEWS[i][1])
+                                .withText((String) REVIEWS[i][2]).withScore((Float) REVIEWS[i][3])));
+                    }
+                }));
+
+    }
+
+    @Test(expected = io.sphere.sdk.client.ErrorResponseException.class)
```
schleichardt: What is the goal of the tests with the invalid products or customers?

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
njacintoct: The SDK is not responsible for processing the reply from the server? if so, how do you know that is working correctly if you don't test what is on the object?
Unless you mean that the creation and update times are not relevant information for the client and if so, why is it being provided?

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
njacintoct: elaborate please

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
njacintoct: didn't knew that. is there a sequence for the versions?

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
njacintoct: I didn't type it every time. Simple copy past of the annotation.

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
njacintoct: removed from intellij

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
njacintoct: Intellij didn't got that memo and still creates the test classes with test in the name, sorry about that

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
schleichardt: What do you mean by sequence?

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
schleichardt: http://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
schleichardt: http://sphereio.github.io/sphere-jvm-sdk/javadoc/master/io/sphere/sdk/meta/ExceptionDocumentation.html

If you provide invalid arguments sphere responds with error code 400 which will be mapped in io.sphere.sdk.client.ExceptionFactory.

So this kind of error code handling is not domain specific and needs not to be tested except, you provide a subclass of io.sphere.sdk.models.SphereError which is part of io.sphere.sdk.client.ErrorResponseException.

schleichardt: In addition just checking the exception class is not sufficient since the exception could be caused by another error source.

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
```diff
@@ -0,0 +1,392 @@
...
+                            client().execute(ReviewUpdateCommand.of(
+                                    Versioned.of(review.getId(), review.getVersion()), Arrays.asList(
+                                            SetTitle.of(Optional.empty()),
+                                            SetText.of(Optional.empty()),
+                                            SetScore.of(Optional.empty()))));
+                        })));
+    }
+
+    @Test
+    public void testCreate_ConstraintTitleTextScore() throws Exception {
```
schleichardt: For me is not clear what this test is about, also it contains a lot of repetition.

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
```diff
@@ -0,0 +1,392 @@
...
+    public void testCreate_NoAuthorName() throws Exception {
+        withProduct(client(), product ->
+                withCustomer(client(), customer -> {
+                    Review review = client().execute(new ReviewCreateCommand(ReviewDraft.of(product.getId(), customer.getId())
+                            .withTitle((String) REVIEWS[0][1])));
+                }));
+    }
+
+    @Test
+    public void testGetAll() throws Exception {
```
schleichardt: I think you need not this test.

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
```diff
@@ -0,0 +1,392 @@
...
+                            assertTrue(optReview.isPresent());
+                            assertEquals(optReview.get(), review);
+                        })));
+    }
+
+    @Test
+    public void testGetById() throws Exception {
+        withProduct(client(), product ->
+                withCustomer(client(), customer ->
+                        withReview(client(), product, customer, review -> {
```
schleichardt: The block

```
        withProduct(client(), product ->
                withCustomer(client(), customer ->
                        withReview(client(), product, customer, review -> {
```
Is repeated a lot, I think it would be nicer if "withReview" takes care of creating a product a customer underneath.

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
```diff
@@ -0,0 +1,392 @@
...
+                            PagedQueryResult<Review> result = client().execute(ReviewQuery.of()
+                                    .withPredicate(Predicate.of("customerId = \"" + review.getCustomerId() + "\"")));
+                            Optional<Review> optReview =
+                                    result.getResults().stream().filter(review1 -> review1.getId().equals(review.getId())).findFirst();
+                            assertTrue(optReview.isPresent());
+                            assertEquals(optReview.get(), review);
+
+                            result = client().execute(ReviewQuery.of()
```
schleichardt: In the JVM SDK we are going a object-functional approach and we don't use reassignment. You may have observed, that we put "final" everywhere.

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
```diff
@@ -0,0 +1,392 @@
...
+                                    .withPredicate(Predicate.of("productId = \"" + review.getProductId() + "\"")));
+                            optReview = result.getResults().stream().filter(review1 -> review1.getId().equals(review.getId())).findFirst();
+                            assertTrue(optReview.isPresent());
+                            assertEquals(optReview.get(), review);
+
+                            result = client().execute(ReviewQuery.of()
+                                    .withPredicate(Predicate.of("productId = \"" + review.getProductId() + "\" and customerId = \"" + review.getCustomerId() + "\"")));
```
schleichardt: I t would be nice if you could here use ReviewQueryModel instead of String concatenation.

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
```diff
@@ -0,0 +1,392 @@
...
+                            Instant startUpdate = Instant.now().minusSeconds(SERVER_TIME_TOLERANCE_SEC); // time tolerance
+                            Review review = client().execute(ReviewUpdateCommand.of(
+                                    Versioned.of(newReview.getId(), newReview.getVersion()), SetTitle.of(Optional.of((String) REVIEWS[1][1]))));
+                            Instant afterUpdate = Instant.now().plusSeconds(SERVER_TIME_TOLERANCE_SEC); // seconds tolerance
+                            assertEquals((String) REVIEWS[1][1], review.getTitle().get());
+                            assertNotEquals(newReview.getVersion(), review.getVersion());
+                            checkTime(review.getLastModifiedAt(), review.getLastModifiedAt(),
+                                    startUpdate, afterUpdate);
+                            review = client().execute(ReviewUpdateCommand.of(
+                                    Versioned.of(review.getId(), review.getVersion()), SetTitle.of(Optional.empty())));
```
schleichardt: Instead of `Versioned.of(review.getId(), review.getVersion())` you can use directly `review`.

models/src/main/java/io/sphere/sdk/reviews/ReviewDraft.java
```diff
@@ -0,0 +1,90 @@
...
+        return ReviewDraftBuilder.of(this).authorName(authorName).build();
+    }
+
+    public ReviewDraft withAuthorName(final String authorName) {
+        return withAuthorName(Optional.of(authorName));
+    }
+
+
+    public ReviewDraft withTitle(final Optional<String> title) {
+        return ReviewDraftBuilder.of(this).title(title).build();
```
schleichardt: You could extract a private method `builder` for `ReviewDraftBuilder.of(this)`.

models/src/main/java/io/sphere/sdk/reviews/commands/ReviewCreateCommand.java
```diff
@@ -0,0 +1,23 @@
+package io.sphere.sdk.reviews.commands;
+
+import io.sphere.sdk.reviews.Review;
+import io.sphere.sdk.reviews.ReviewDraft;
+import io.sphere.sdk.commands.CreateCommandImpl;
+
+/** Creates a review.
```
schleichardt: Thanks for the documentation.

models/src/main/java/io/sphere/sdk/reviews/commands/ReviewUpdateCommand.java
```diff
@@ -0,0 +1,35 @@
...
+import io.sphere.sdk.commands.CreateCommandImpl;
+
+import java.util.List;
+
+import static java.util.Arrays.asList;
+
+/** Creates a review.
+
+ <p>Example:</p>
+ {@include.example io.sphere.sdk.reviews.ReviewIntegrationTest#createReview()}
```
schleichardt: The link should go to an update example.

models/src/main/java/io/sphere/sdk/reviews/commands/ReviewUpdateCommand.java
```diff
@@ -0,0 +1,35 @@
...
+
+@see io.sphere.sdk.reviews.ReviewDraftBuilder
```
schleichardt: Can be omitted.

models/src/main/java/io/sphere/sdk/reviews/commands/updateCommands/SetAuthorName.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+public class SetAuthorName extends UpdateAction<Review> {
+    private final Optional<String> authorName;
+
+    private SetAuthorName(final Optional<String> authorName) {
+        super("setAuthorName");
+        this.authorName = authorName;
+    }
+
+    public static SetAuthorName of(final Optional<String> authorName) {
```
schleichardt: Could you also provide a convenience method 

```
public static SetAuthorName of(final String authorName) {
    return of(Optional.of(authorName));
}
```

models/src/main/java/io/sphere/sdk/reviews/commands/updateCommands/SetScore.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+public class SetScore extends UpdateAction<Review> {
+    private final Optional<Float> score;
+
+    private SetScore(final Optional<Float> score) {
+        super("setScore");
+        this.score = score;
+    }
+
+    public static SetScore of(final Optional<Float> score) {
```
schleichardt: please provide a convenience method for `final float score`.

models/src/main/java/io/sphere/sdk/reviews/commands/updateCommands/SetText.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+public class SetText extends UpdateAction<Review> {
+    private final Optional<String> text;
+
+    private SetText(final Optional<String> text) {
+        super("setText");
+        this.text = text;
+    }
+
+    public static SetText of(final Optional<String> text) {
```
schleichardt: convenience method ;)

models/src/main/java/io/sphere/sdk/reviews/commands/updateCommands/SetTitle.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+public class SetTitle extends UpdateAction<Review> {
+    private final Optional<String> title;
+
+    private SetTitle(final Optional<String> title) {
+        super("setTitle");
+        this.title = title;
+    }
+
+    public static SetTitle of(final Optional<String> title) {
```
schleichardt: ;)

models/src/test/java/io/sphere/sdk/reviews/ReviewDraftBuilderTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+import org.junit.Test;
+
+import java.util.Optional;
+
+import static org.junit.Assert.*;
+
+/**
+ * Created by njacinto on 21/04/15.
+ */
```
schleichardt: please remove

models/src/test/java/io/sphere/sdk/reviews/ReviewDraftTest.java
```diff
@@ -0,0 +1,142 @@
...
+import static org.junit.Assert.*;
+
+/**
+ * Created by njacinto on 21/04/15.
+ */
+public class ReviewDraftTest {
+    static final ReviewDraft DEFAULT_REVIEW;
+
+
+    static {
```
schleichardt: Why not (with appropriate newlines)

```
private static final ReviewDraft DEFAULT_REVIEW = ReviewDraft.of(ReviewImplTest.DEFAULT_PRODUCT_ID, ReviewImplTest.DEFAULT_CUSTOMER_ID).withAuthorName(ReviewImplTest.DEFAULT_AUTHOR_NAME).withTitle(ReviewImplTest.DEFAULT_TITLE).withText(ReviewImplTest.DEFAULT_TEXT).withScore(ReviewImplTest.DEFAULT_SCORE);
```

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
njacintoct: sorry, I miss understood the reason for the comment. I will remove the test.
Concerning your last comment, there are many reasons why the tests will fail, that are not being handle in every test. Hopefully this are cover and other tests that will also fail, otherwise it will be the same has testing everything on all the tests.

models/src/test/java/io/sphere/sdk/reviews/ReviewDraftTest.java
schleichardt: With fest assertions we would have `assertThat(review).isNotSameAs(reviewWith);`

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
njacintoct: to see if it fails. isn't that what is expected?

models/src/test/java/io/sphere/sdk/reviews/ReviewImplTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+import java.time.Instant;
+import java.util.Optional;
+
+import static org.junit.Assert.*;
+
+/**
+ * Created by njacinto on 21/04/15.
+ */
+public class ReviewImplTest {
```
schleichardt: I would skip this test, since we should test in a ReviewCreateCommand if all getters work as expected and match the draft object.

models/src/test/java/io/sphere/sdk/reviews/commands/ReviewsEndpointTest.java
```diff
@@ -0,0 +1,20 @@
...
+import io.sphere.sdk.reviews.Review;
+import io.sphere.sdk.reviews.queries.*;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+/**
+ * Created by njacinto on 21/04/15.
+ */
+public class ReviewsEndpointTest {
```
schleichardt: This is implicitly tested by the integration tests. So you can delete this class.

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
njacintoct: is testing if is possible to create reviews with several combinations of title, score and text

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
njacintoct: the get all?

models/src/test/java/io/sphere/sdk/reviews/commands/updateCommands/SetAuthorNameTest.java
```diff
@@ -0,0 +1,20 @@
...
+import org.junit.Test;
+
+import java.util.Optional;
+
+import static org.junit.Assert.*;
+
+/**
+ * Created by njacinto on 21/04/15.
+ */
+public class SetAuthorNameTest {
```
schleichardt: This class is implicitly tested with the ReviewUpdateCommand test. So we can get rid of this test.

schleichardt: And others like this.

models/src/test/java/io/sphere/sdk/reviews/queries/ReviewQueryModelTest.java
```diff
@@ -0,0 +1,31 @@
...
+import io.sphere.sdk.queries.StringQuerySortingModel;
+import io.sphere.sdk.reviews.Review;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+/**
+ * Created by njacinto on 21/04/15.
+ */
+public class ReviewQueryModelTest {
```
schleichardt: I think this class is also obsolete since the integration tests count.

You should know, that a query is not for all fields possible and you need to verify if it is possible with an integration test.

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
njacintoct: not all numbers on the code have to be constants. It's used, in that context, once.

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
njacintoct: can I expect a certain number or just compare has different?

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
schleichardt: I meant the english word exception (like "every can is open except one") not Javas Exception class.

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
schleichardt: It belongs to code smells. If anyone just jumps into this test it is not clear what is meant by 2.

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
schleichardt: But you are then testing the API and the SDK. You should only test the SDK.

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
schleichardt: I think one test with leaving out all optional values and one test with all possible filled fields is sufficient.

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
schleichardt: Yes, you are testing the API, not the SDK.

models/src/test/java/io/sphere/sdk/reviews/commands/updateCommands/SetAuthorNameTest.java
njacintoct: you mean the integration test?

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
schleichardt: ReviewImpl extends class io.sphere.sdk.models.DefaultModelImpl which contains the fields `final String id, final long version, final Instant createdAt, final Instant lastModifiedAt`, which are already tested, so you don't need to worry about the fields since they are tested elsewhere.

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
hajoeichler: @schleichardt can you please explain why you don't need this test case? Is it covered by something else?

models/src/it/java/io/sphere/sdk/reviews/ReviewIntegrationTest.java
hajoeichler: http://www.adam-bien.com/roller/abien/entry/are_naming_conventions_for_junit

PR #374, opened 2015-04-24
SCLIENTS-302: add validFrom and validUntil to product price

yanns:


yanns:
@schleichardt what is your strategy to test this kind of changes?

schleichardt:
You could add a method io.sphere.sdk.products.commands.ProductUpdateCommandTest#addPriceWithValidityPeriod where you add a price with both values to a product and check if it is then contained in the product.

yanns:
Thx!
I have one problem with the serializer:
the `LocalDateTime` is serialized into `"validFrom":[2015,4,25,10,9,13,148000000]`

I sae JsonUtils#newObjectMapper registers a serializer for `LocalDateTime`.
Maybe it is not used in the tests? Can you help here?

schleichardt:
Hmm, strange that there is a serializer and it works not as expected. 

Why is it LocalDateTime? For temporal data with a specific day in the year and specific hour in the day I use Instant: https://github.com/sphereio/sphere-jvm-sdk/blob/master/sphere-common/src/main/java/io/sphere/sdk/models/DefaultModel.java#L19

But maybe it is not the correct type anyway.

schleichardt:
https://github.com/sphereio/sphere-jvm-sdk/issues/375

lauraluiz:
Are these datetimes converted to UTC?

From http://java.dzone.com/articles/java-8-apis-javautiltime:
LocalDateTime: a date and the time within that date, no time zone.
Instant: point in time without time zone.
"But there is a difference: LocalDateTime is not a point on the time line as Instant is, LocalDateTime is just a date and time as a person would write on a note. Consider the following example: two persons which were born at 11am, July the 2nd 2013. The first was born in the UK while the second in California. If we ask any of them for their birth date it will look that they were born on the same time (this is the LocalDateTime) but if we align the dates on the timeline (using Instant) we will find out that the one born in California is few hours younger than the one born in the UK (NB: to create the appropriate Instant we have to convert the time to UTC, this is where the difference lays)"

yanns:
I changed the implementation to use `Instant`.
About the discussion about time zone or not, I think it concerns much more than the SDK.
The API itself is not clear in my opinion. We accept date time with a time zone (http://dev.sphere.io/http-api-types.html#datetime), but all date times are converted to UTC.
Using `Instant` in the SDK makes clear that the API treats all data time in UTC.
If the client wants to use `LocalDateTime`, they just have to call `toInstant` to store the validity period.

schleichardt:
Thanks Yann! If we need to change the date type this does not affect this pull request.

yanns:
The PR is now opened for review.

schleichardt:
Thank you very much!

PR #380, opened 2015-04-27
Fix ProductAllVariantsQueryModel is package scope and therefore not usable

schleichardt:


lauraluiz:
Good catch :disappointed_relieved:

schleichardt:
I wondered why it even compiled.

Another example of: "It works in tests, but not in production."

lauraluiz:
Actually that's the evil side of package scope.

schleichardt:
It is not package scopes fault, if the tests for a public API would not be in the same package then you discover this kind of problems.

Can I merge?

lauraluiz:
Yes, merge it.

PR #384, opened 2015-04-28
Update library dependencies.

schleichardt:


build.sbt
```diff
@@ -28,8 +28,6 @@ autoScalaLibrary in ThisBuild := false // no dependency to Scala standard librar
 
 crossPaths in ThisBuild := false
 
-Release.publishSettings
```
lauraluiz: What did it do?

lauraluiz:
Good.

lauraluiz:
But check the build error.

build.sbt
schleichardt: It sets the publish settings to sonatype, but was duplicated.

schleichardt: see line 49/47

PR #386, opened 2015-04-28
Add additional check to avoid flaky tests.

lauraluiz:
So the problem is that there is a delay from the moment the products are in ES and the moment when they can be used for some facets, sort and filter requests. I solved it with an additional check that seems to work very well in localhost. Anyway, I've increased the delay just to make sure the check is not skipping some case.

Please @schleichardt review.

Fixes https://github.com/sphereio/sphere-jvm-sdk/issues/382

### DIFF
--- a/sphere-models/src/it/java/io/sphere/sdk/products/ProductProjectionSearchIntegrationTest.java
+++ b/sphere-models/src/it/java/io/sphere/sdk/products/ProductProjectionSearchIntegrationTest.java
@@ -61,14 +61,17 @@ public class ProductProjectionSearchIntegrationTest extends IntegrationTest {
         setupTestProducts();
         setupEvilTestProducts();
         final SearchSort<ProductProjection> sortByCreatedAt = ProductProjectionSearch.model().createdAt().sort(SimpleSearchSortDirection.DESC);
-        final SearchDsl<ProductProjection> search = ProductProjectionSearch.of(STAGED).withSort(sortByCreatedAt);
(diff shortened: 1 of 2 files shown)


sphere-models/src/it/java/io/sphere/sdk/products/ProductProjectionSearchIntegrationTest.java
```diff
@@ -61,14 +61,17 @@ public static void setupProducts() {
...
-        final SearchDsl<ProductProjection> search = ProductProjectionSearch.of(STAGED).withSort(sortByCreatedAt);
+        final TermFacetExpression<ProductProjection, String> facet = ProductProjectionSearch.model().allVariants().attribute().ofText(ATTR_NAME_COLOR).facetOf().all();
+        final SearchDsl<ProductProjection> search = ProductProjectionSearch.of(STAGED).withSort(sortByCreatedAt).plusFacet(facet);
         execute(search, res -> {
             final List<String> ids = SphereTestUtils.toIds(res.getResults());
-            return ids.contains(product1.getId()) && ids.contains(product2.getId()) && ids.contains(product3.getId())
-                    && ids.contains(evilProduct1.getId()) && ids.contains(evilProduct2.getId());
+            final boolean productsExist = ids.contains(product1.getId()) && ids.contains(product2.getId())
+                    && ids.contains(product3.getId()) && ids.contains(evilProduct1.getId()) && ids.contains(evilProduct2.getId());
+            final boolean facetsExist = res.getTermFacetResult(facet).map(term -> term.getMissing() > 1 && term.getTotal() > 1).orElse(false);
```
schleichardt: I don't understand yet, why the facet check helps or is necessary.

PR #389, opened 2015-04-28
Missing order things 159

schleichardt:
Not complete, but if we merge now we have only a few changes.

@lauraluiz please review.


lauraluiz:
I've reviewed everything, we can merge (once my comment is solved).

Issue #397, opened 2015-05-06
[1.0.0-M13] CustomerSignInCommand returns empty ErrorResponseException


MeiSign:
We receive an empty ErrorResponseException when we run a CustomerSignInCommand. From what I can say we use the command correctly and use a valid email address and password as  parameters but it is hard to debug because of the cryptic exception.
Is it possible that this command is broken since one of the last releases? We think that this command did work and we have not changed our code except updating the sphere dependencies.

Code:

    def sphereSignIn(credentials: UserCredentials): Future[Boolean] = {
        println(s"${credentials.email.value}:${credentials.password.value}")
        val signInQuery = CustomerSignInCommand.of(
            credentials.email.value, credentials.password.value)

        sphereClient.execute(signInQuery)
            .map(_ => true)
            .recover {
              case e: Exception =>
                throw e
                false 
            }
    }

 Exception:

    [error] play - Cannot invoke the action, eventually got an error: java.util.concurrent.CompletionException: io.sphere.sdk.client.ErrorResponseException: SDK: 1.0.0-M13
    project: <unknown>
    Java: 1.8.0_45
    cwd: /Users/Stefan/Workspace/Scala/neeedo-api
    date: Wed May 06 23:09:48 CEST 2015
    sphere request: <unknown>
    http request: <unknown>
    http response: <unknown>
    additional notes: []
    Javadoc: http://sphereio.github.io/sphere-jvm-sdk/javadoc/v1.0.0-M13/io/sphere/sdk/client/ErrorResponseException.html

    [error] application -  ! @6m442ffmf - Internal server error, for (POST) [/offers] ->

    play.api.Application$$anon$1: Execution exception[[CompletionException: io.sphere.sdk.client.ErrorResponseException: SDK: 1.0.0-M13
    project: <unknown>
    Java: 1.8.0_45
    cwd: /Users/Stefan/Workspace/Scala/neeedo-api
    date: Wed May 06 23:09:48 CEST 2015
    sphere request: <unknown>
    http request: <unknown>
    http response: <unknown>
    additional notes: []
    Javadoc: http://sphereio.github.io/sphere-jvm-sdk/javadoc/v1.0.0-M13/io/sphere/sdk/client/ErrorResponseException.html
    ]]
	    at play.api.Application$class.handleError(Application.scala:296) ~[play_2.11-2.3.8.jar:2.3.8]
[...]

PR #399, opened 2015-05-08
Improve error messages and documentation for CustomerSignInCommand. #397

schleichardt:
@lauraluiz please review

The CustomerSignInCommand throwed in the error case an exception without the request and the response information.
I refactored a little bit the client impl to make the stack trace nicer.


lauraluiz:
Reviewed as well, merging time.

PR #405, opened 2015-05-11
Project endpoint 364

schleichardt:
@lauraluiz please review

### DIFF
--- a/sphere-models/src/main/java/io/sphere/sdk/projects/Project.java
+++ b/sphere-models/src/main/java/io/sphere/sdk/projects/Project.java
@@ -2,16 +2,21 @@ package io.sphere.sdk.projects;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.core.type.TypeReference;
+import com.neovisionaries.i18n.CountryCode;
 import io.sphere.sdk.models.Base;
+
+import java.util.List;
 
 public final class Project extends Base {
     private final String key;
     private final String name;
+    private final List<CountryCode> countries;
 
     @JsonCreator
-    private Project(final String key, final String name) {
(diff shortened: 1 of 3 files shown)


sphere-models/src/main/java/io/sphere/sdk/projects/Project.java
```diff
@@ -2,20 +2,56 @@
...
+
+    public List<CountryCode> getCountries() {
+        return countries;
+    }
+
+    public List<String> getLanguages() {
+        return languages;
+    }
+
+    public String getTrialUntil() {
```
schleichardt: Or would you prefer here a special data container for year and month?

```json
"trialUntil" : "2014-11"
```

lauraluiz: Hmm... a tricky date format. This is usable right now, but anybody that wants to use it will have to convert it to something usable. Either in this PR or on a future one we have to implement a special data container IMO.

sphere-models/src/main/java/io/sphere/sdk/projects/Project.java
```diff
@@ -2,20 +2,58 @@
...
+
+    public List<CountryCode> getCountries() {
+        return countries;
+    }
+
+    public List<String> getLanguages() {
+        return languages;
+    }
+
+    public ZonedDateTime getTrialUntil() {
```
schleichardt: OK? @lauraluiz

lauraluiz: Yes, probably that's the best way to go. Technically the trial ends up in a particular second, even if the month/year format doesn't explicitly mention it.

lauraluiz:
We can merge it.

PR #406, opened 2015-05-11
Rename SDK Predicate to QueryPredicate. close #401

### LINKED ISSUE #401: Rename SDK Predicate to QueryPredicate
The name clashes with the JDK class name, ok therefore we have packages, but we also have QuerySort so there is kind of a symmetry.

@lauraluiz What do you think?

schleichardt:
@lauraluiz please review


sphere-common/src/main/java/io/sphere/sdk/queries/IsGreaterThanOrEqualsQueryPredicate.java
```diff
@@ -0,0 +1,13 @@
+package io.sphere.sdk.queries;
+
+class IsGreaterThanOrEqualsQueryPredicate<T, V, M> extends ComparisonQueryPredicate<T, V, M> {
+
+    public IsGreaterThanOrEqualsQueryPredicate(final QueryModel<M> queryModel, final V value) {
```
lauraluiz: Now that we are renaming these predicates, a small discussion regarding the naming:
"price is greater than 3" or "price equals 4" I think are both correct (not sure how natural the second one is in this sentence, though).
But I think you would always say "price is greater than or equal to 4". At least that's the reason why I wrote it like this in the Search API: https://github.com/sphereio/sphere-jvm-sdk/blob/master/sphere-common/src/main/java/io/sphere/sdk/search/RangedFacetSearchModel.java#L54

So from my humble knowledge of English grammar I would use the "equal to" version because that's what I found people use when talking. What version is more normal to you?

schleichardt: This class is package scope and my gets removed due to duplication. We can later on discuss together the naming: https://github.com/sphereio/sphere-jvm-sdk/issues/407

But this is not related to this pr.

lauraluiz: The class name was an example. But all through the Query API this kind of naming is used for public methods as well. What I would like is to solve this issue before the release, to avoid additional breaking changes in the Query API in the next release.

schleichardt: Ok, I put https://github.com/sphereio/sphere-jvm-sdk/issues/407 into the current milestone.

schleichardt:
@lauraluiz So can I merge it? The method naming thing will be in another pull request.

lauraluiz:
Yes, the rest is reviewed, so go ahead.

PR #441, opened 2015-05-26
Add VariantIdentifier getter to product variant. This makes it possib…

### LINKED ISSUE #239: Improve VariantIdentifier
Product and ProductProjection could  be optimized in the constructor so they contain variant which can return the variant identifiers.

Also it would be nice to use variant identifier as parameter where possible.

schleichardt:
…le to get the product id from the variant. close #239 
@lauraluiz please review, but this is a pr for a later milestone.

### DIFF
--- a/sphere-models/src/main/java/io/sphere/sdk/products/ProductVariant.java
+++ b/sphere-models/src/main/java/io/sphere/sdk/products/ProductVariant.java
@@ -25,7 +25,9 @@ public interface ProductVariant extends AttributeContainer {
     Optional<ProductVariantAvailability> getAvailability();
 
     /**
-     * Gets the id of the product and the variant. This operation is may not available.
+     * Gets the id of the product and the variant. This operation may not be available.
+     * It will be available if this {@link ProductVariant} has been created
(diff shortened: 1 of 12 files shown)


sphere-models/src/main/java/io/sphere/sdk/products/ProductVariant.java
```diff
@@ -25,4 +23,11 @@
      * @return availability
      */
     Optional<ProductVariantAvailability> getAvailability();
+
+    /**
+     * Gets the id of the product and the variant. This operation is may not available.
```
lauraluiz: *may not be available

sphere-models/src/main/java/io/sphere/sdk/products/ProductVariant.java
```diff
@@ -25,4 +23,11 @@
...
+     * @return identifier for this variant
+     * @throws UnsupportedOperationException if the operation is not available
```
lauraluiz: We need to explain when this operation may not be available.

lauraluiz:
It can be merged after fixing the comments.
This PR will save a lot of lines of code in the client app ;)

PR #444, opened 2015-05-26
Use AutoCloseable instead of Closeable for SphereClient. close #163

### LINKED ISSUE #163: Please use AutoCloseable instead of Closeable for JavaClient
Currently the JavaClient implements Closeable an interface used for IO on streams, according to the docs:
```
A Closeable is a source or destination of data that can be closed.
The close method is invoked to release resources that the object is holding
(such as open files).
...
Closes this stream and releases any system resources associated
with it. If the stream is already closed then invoking this method has no effect.
```
Most likely the AutoCloseable interface is what matches more, as this is intended for arbitrary resources:
```
A resource that must be closed when it is no longer needed.
```

schleichardt:


lauraluiz:
:+1:

PR #445, opened 2015-05-26
Document generics noteworthiness with CustomObjectQuery. close #295

### LINKED ISSUE #295: Improve generics with CustomObjectQuery
`CustomObjectQuery.model().id().sort(SortDirection.ASC)` creates `Sort<CustomObject<Object>>`.

It would be cooler to not to need to pass the type as in here:

`final Sort<CustomObject<JsonNode>> sort = CustomObjectQuery.<JsonNode>model().id().sort(SortDirection.ASC);`

schleichardt:


src/main/java/io/sphere/sdk/meta/CustomObjectDocumentation.java
```diff
@@ -148,6 +150,12 @@ Instead of pure JSON it is encouraged to use Java objects (POJO) with the Jackso
 
  {@include.example io.sphere.sdk.customobjects.migrations.CustomObjectsMigrationsTest#exampleForMigrationCall()}
 
+
+ <h3 id="traps">Traps</h3>
+
+ <p>Unlike other query models, {@link CustomObjectQuery#model()} takes a type parameter of the result type of the custom object.</p>
```
lauraluiz: I would continue the sentence with something that focuses the attention on the type parameter declaration, to make sure they understand what we mean with that. Something like "Notice that it is necessary to explicitly declare the type when requesting the query model, as shown in the following examples" (better phrased).

PR #468, opened 2015-06-10
More reference expansion

schleichardt:
@lauraluiz not completely done, but it includes the breaking changes and is already very big.


sphere-common/src/main/java/io/sphere/sdk/queries/MetaModelExpansionDsl.java
```diff
@@ -0,0 +1,27 @@
+package io.sphere.sdk.queries;
+
+import java.util.function.Function;
+
+public interface MetaModelExpansionDsl<T, C, E> {
```
lauraluiz: Missing explanation of T, C, E

sphere-common/src/main/java/io/sphere/sdk/queries/MetaModelFetchDsl.java
```diff
@@ -0,0 +1,25 @@
+package io.sphere.sdk.queries;
+
+import java.util.List;
+import java.util.function.Function;
+
+public interface MetaModelFetchDsl<T, C extends MetaModelFetchDsl<T, C, E>, E> extends FetchDsl<T, C>, MetaModelExpansionDsl<T, C, E> {
```
lauraluiz: Missing explanation of T, C, E

sphere-models/src/it/java/io/sphere/sdk/categories/queries/CategoryPredicatesTest.java
```diff
@@ -21,7 +21,7 @@
 public class CategoryPredicatesTest extends IntegrationTest {
     @Test
     public void isLessThanComparisonPredicate() throws Exception {
-        final QueryPredicate<Category> predicate = CategoryQuery.model().name().lang(Locale.ENGLISH).isLessThan("2");
+        final QueryPredicate<Category> predicate = CategoryQueryModel.of().name().lang(Locale.ENGLISH).isLessThan("2");
```
lauraluiz: Change to lambda, along with all the other predicates in this test.

lauraluiz:
Reviewed :sweat_smile:

sphere-models/src/it/java/io/sphere/sdk/categories/queries/CategoryPredicatesTest.java
schleichardt: It is better to keep is with the value shape since I use it as parameter to a function.

PR #472, opened 2015-06-11
Add product type in search endpoint for products

lauraluiz:
@schleichardt please review.

### DIFF
--- a/sphere-models/src/main/java/io/sphere/sdk/products/search/ExperimentalProductProjectionSearchModel.java
+++ b/sphere-models/src/main/java/io/sphere/sdk/products/search/ExperimentalProductProjectionSearchModel.java
@@ -2,6 +2,7 @@ package io.sphere.sdk.products.search;
 
 import io.sphere.sdk.categories.Category;
 import io.sphere.sdk.products.ProductProjection;
+import io.sphere.sdk.producttypes.ProductType;
 import io.sphere.sdk.search.*;
 
 import java.util.Optional;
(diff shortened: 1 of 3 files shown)


schleichardt:
I would prefer to have an integration test.

lauraluiz:
Actually I was about to write a task to write a IT per element (i.e. categories, name, createdAt, modifiedAt, price, one per attribute...). I don't see the point of writing it just for one. But that means 22 tests * 3 (filter/facet/sort). And quite more when sets are available.
In the UT I'm making sure the generated expression is exactly what the API documentation requires. Implementing IT for each type of expression is kind of testing the API works correctly.
So, for coherence, I don't think we should implement IT for this. Or, in any case, I would implement only for references (categories, productType) since we have the additional possible error of the name.

lauraluiz:
Just if you were wondering, the list of tests should match these: https://github.com/sphereio/sphere-jvm-sdk/blob/master/sphere-models/src/test/java/io/sphere/sdk/products/search/ProductProjectionSearchTest.java

schleichardt:
If you will create tests for those you can merge.

Yes, it is not the goal to test the API, but often enough integration tests showed, that a feature was not accordingly implemented in the SDK or backend or not present at all. Or I found behaviour which was strange because of undocumented special cases.

lauraluiz:
Yep, I would love to have a IT for each situation to make sure nothing breaks the JVM SDK. But I think it's quite unrealistic to be able to do it.
I've created an issue to decide the best approach: https://github.com/sphereio/sphere-jvm-sdk/issues/473

schleichardt:
@lauraluiz Can you merge this?

PR #486, opened 2015-06-16
SDK http library

schleichardt:
The HTTP lib and client implementations for other SDKs to reuse.

@lauraluiz 
please review, but it is not urgent


lauraluiz:
Good! You can merge.

PR #511, opened 2015-06-26
Improve instantiation of AHC.

schleichardt:


hajoeichler:
So much time for so few lines ;)

PR #601, opened 2015-07-26
Api/512/nullable instead of optional

schleichardt:


hajoeichler:
screen shot 2015-07-26 at 22 36 39 :facepunch:

schleichardt:
:relieved:

PR #646, opened 2015-08-11
Add longer example for creating api objects in unit tests. close #645

### LINKED ISSUE #645: Add longer example for creating api objects in unit tests

schleichardt:


schleichardt:
@lauraluiz please review.

A preview is on http://sphereio.github.io/sphere-jvm-sdk/javadoc/documentation/645/test-double-documentation/io/sphere/sdk/meta/TestingDocumentation.html starting with "Detailed example for creating test doubles".

schleichardt:
I think I should put a link to the impex playground.

src/test/resources/product-projection.json
```diff
@@ -0,0 +1,56 @@
...
+    "images": [
+      {
+        "url": "https://www.commercetools.com/cli/data/253234387_1.jpg",
+        "dimensions": {
+          "w": 1400,
+          "h": 1400
+        }
+      }
+    ],
+    "attributes": []
```
butenkor: Would be cool if examples and test were a bit more realistic and had attributes like 99% of projects does.

schleichardt: Sure, where can I get test data with realistic product attributes?

butenkor: Add more variants and define couple self explanatory attributes like ```brand``` of type text, ```colour``` and ```size``` of type lenum.

lauraluiz: Actually I like that the product data used is the one we are giving as example data, so people can really reproduce it. I don't see a strong reason to need attributes there just to show how to test data. In any case we should add attributes and more realistic data to the example data we offer.

schleichardt: The thing is, is it good enough for now? There will be a time to improve the documentation but now we need to get RC1 done. This docu is just a side product of a support question.

lauraluiz: IMHO the more complex the data, the more difficult it is to understand. I would only show a simple product. It's not like making it more complex is going to affect the test. Also, people don't have to care about the complexity because they just have to copy and paste stuff from impex, not writing it themselves.

schleichardt:
@lauraluiz Is this good enough (to merge) for now?

src/test/resources/product-projection.json
butenkor: It is not going to affect the test but it should show best practice how you can easily test without newly removed builders. Current example is definitely better then nothing but in reality you always have to test much more: variants and its attributes (different kinds and amount), prices, references. Currently M16 SDK user has to discover on its own how to reuse given JSON example in order to test 20 different products or product search outcomes without defining multiple JSON files.

One simple example and one complex example would be nice.

lauraluiz:
Yes, but let's create a task to make a very simple and complex example for the future before.

PR #647, opened 2015-08-11
Enable path search

lauraluiz:
@schleichardt please review.

After many tries to fix the facet types, I gave up on that (after adding many IT, as a positive aspect).
But the important feature from this PR is that now you are able to access the path from a Facet or Filter expression. This is important to build the complete functionality from a Facet UI.


sphere-common/src/main/java/io/sphere/sdk/search/TermFacetExpression.java
```diff
@@ -5,6 +5,7 @@
 import static java.util.Collections.emptyList;
 
 public class TermFacetExpression<T, V> extends TermExpression<T, V> implements FacetExpressionBase<T> {
+    SearchModel<T> model;
```
schleichardt: you give the search model to the super class constructor, why do you add a field in the subclass which looks unused?

sphere-models/src/it/java/io/sphere/sdk/products/ProductProjectionSearchFilterTypesIntegrationTest.java
```diff
@@ -102,15 +105,24 @@ public static void setupProducts() {
...
     }
 
+    @Ignore
     @Test
-    public void filtersTextAttributes() throws Exception {
+    public void booleanAttributesFacets() throws Exception {
+        final TermFacetExpression<ProductProjection, Boolean> facetExpr = model().allVariants().attribute().ofBoolean(ATTR_NAME_BOOLEAN)
+                .faceted().byAllTerms();
+        final ProductProjectionSearch search = ProductProjectionSearch.ofStaged().withFacets(facetExpr);
+        //assertThat(executeAndReturnTerms(search, facetExpr)).containsOnlyElementsOf(asList(TermStats.of(true, 1), TermStats.of(false, 1))); // still returning F, T. to be fixed by API
```
schleichardt: Is there a JVM SDK issue with the label HTTP API related?

schleichardt: Could we provide a workaround which works with true/false and "F"/"T" so that we don't need to adjust it later?

sphere-models/src/it/java/io/sphere/sdk/products/ProductProjectionSearchFilterTypesIntegrationTest.java
```diff
@@ -158,54 +210,121 @@ public void filtersLocEnumLabelAttributes() throws Exception {
...
         assertThat(executeAndReturnIds(rangeSearch)).containsOnly(product1.getId());
     }
 
+    @Ignore
+    @Test
+    public void numberAttributesFacets() throws Exception {
+        final TermFacetExpression<ProductProjection, BigDecimal> termExpr = model().allVariants().attribute().ofNumber(ATTR_NAME_NUMBER)
+                .faceted().byAllTerms();
+        final ProductProjectionSearch termSearch = ProductProjectionSearch.ofStaged().withFacets(termExpr);
+        //assertThat(executeAndReturnTerms(termSearch, termExpr)).containsOnlyElementsOf(asList(TermStats.ofNumber(valueOf(5D), 1), TermStats.ofNumber(valueOf(10D), 1)));
```
schleichardt: Why is this outcommented?

sphere-models/src/it/java/io/sphere/sdk/products/ProductProjectionSearchFilterTypesIntegrationTest.java
schleichardt: So you could link the issue here.

sphere-models/src/it/java/io/sphere/sdk/products/ProductProjectionSearchFilterTypesIntegrationTest.java
```diff
@@ -346,11 +559,12 @@ public void filtersTimeSetAttributes() throws Exception {
     }
 
     @Test
-    public void filtersDateTimeSetAttributes() throws Exception {
+    public void dateTimeSetAttributesFilters() throws Exception {
         final ProductProjectionSearch termSearch = ProductProjectionSearch.ofStaged()
                 .withQueryFilters(model -> model.allVariants().attribute().ofDateTime(ATTR_NAME_DATE_TIME_SET)
                         .filtered().by(ZonedDateTime.parse("2002-10-12T23:06:10.204+00:00")));
```
schleichardt: Is the same value hardcoded for product1?

schleichardt:
I would say it looks good, but I had some questions.

sphere-models/src/it/java/io/sphere/sdk/products/ProductProjectionSearchFilterTypesIntegrationTest.java
lauraluiz: Because currently we only allow support for String terms.

schleichardt: Is WE SDK or API?

sphere-models/src/it/java/io/sphere/sdk/products/ProductProjectionSearchFilterTypesIntegrationTest.java
lauraluiz: Not sure if I understand you. Do you mean that we can reuse the `ZonedDateTime.parse("2002-10-12T23:06:10.204+00:00") `?

sphere-models/src/it/java/io/sphere/sdk/products/ProductProjectionSearchFilterTypesIntegrationTest.java
lauraluiz: The problem is not the F/T, the problem is that we don't support anything but String terms.

sphere-models/src/it/java/io/sphere/sdk/products/ProductProjectionSearchFilterTypesIntegrationTest.java
lauraluiz: We, SDK. That's what I've been trying to fix for the last two days, with no success. I decided to leave it as it was and wait for the fixes from the API to have a new perspective.

sphere-models/src/it/java/io/sphere/sdk/products/ProductProjectionSearchFilterTypesIntegrationTest.java
schleichardt: I don't understand why this filter works/ what it should do.

lauraluiz: Product 1 has [2001-09-11T22:05:09.203+00:00, 2002-10-12T23:06:10.204+00:00]
Product 2 has [2001-09-11T22:05:09.203+00:00]
If we filter by 2002-10-12T23:06:10.204+00:00 only Product 1 should be returned.

sphere-common/src/main/java/io/sphere/sdk/search/TermFacetExpression.java
lauraluiz: Just I mistake. I added to do some tests and I forgot to remove it. Good eye ;)

sphere-models/src/it/java/io/sphere/sdk/products/ProductProjectionSearchFilterTypesIntegrationTest.java
schleichardt: Could you extract a final variable to make this onvious, for me it was a magic value.

lauraluiz: So we go back to the first interpretation I did from your comment ;)
I will extract this and other complex values. all of them will achieve the opposite goal, I think.

lauraluiz:
@schleichardt can you do a second quick review to the commits I did yesterday? It is about the things we mentioned during the meeting, basically.

schleichardt:
Ok, I am done with review. Only the thing with the centAmount and BigDecimal looksweird.

lauraluiz:
Okay, I will change it to Long. It actually makes things easier :)

PR #649, opened 2015-08-12
Feature/633/product published messages

schleichardt:
@lauraluiz please review

### DIFF
--- a/sphere-models/src/it/java/io/sphere/sdk/products/messages/ProductPublishedMessageTest.java
+++ b/sphere-models/src/it/java/io/sphere/sdk/products/messages/ProductPublishedMessageTest.java
@@ -1,0 +1,31 @@
+package io.sphere.sdk.products.messages;
+
+import io.sphere.sdk.messages.queries.MessageQuery;
+import io.sphere.sdk.products.Product;
+import io.sphere.sdk.products.ProductFixtures;
+import io.sphere.sdk.products.commands.ProductUpdateCommand;
+import io.sphere.sdk.products.commands.updateactions.Publish;
+import io.sphere.sdk.queries.PagedQueryResult;
+import io.sphere.sdk.test.IntegrationTest;
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ProductPublishedMessageTest extends IntegrationTest {
+    @Test
+    public void message() {
+        ProductFixtures.withUpdateableProduct(client(), product -> {
+            assertThat(product.getMasterData().isPublished()).isFalse();
+            final Product publishedProduct = execute(ProductUpdateCommand.of(product, Publish.of()));
+            final PagedQueryResult<ProductPublishedMessage> queryResult = execute(MessageQuery.of()
+                    .withSort(m -> m.createdAt().sort().desc())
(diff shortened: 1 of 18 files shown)


sphere-models/src/it/java/io/sphere/sdk/products/messages/ProductPublishedMessageTest.java
```diff
@@ -0,0 +1,32 @@
...
+    @Test
+    public void message() {
+        ProductFixtures.withUpdateableProduct(client(), product -> {
+            assertThat(product.getMasterData().isPublished()).isFalse();
+            final Product publishedProduct = execute(ProductUpdateCommand.of(product, Publish.of()));
+            final PagedQueryResult<ProductPublishedMessage> queryResult = execute(MessageQuery.of()
+                    .withSort(m -> m.createdAt().sort().desc())
+                    .withExpansionPaths(m -> m.resource())
+                    .withLimit(1)
+                    .forMessageType(ProductPublishedMessage.MESSAGE_HINT));
```
lauraluiz: Since here you are naming the method `forMessageType`, it doesn't match with the constant `MESSAGE_HINT`. I think it would be better to call both either MessageType or MessageHint.

sphere-models/src/main/java/io/sphere/sdk/products/messages/ProductPublishedMessage.java
```diff
@@ -0,0 +1,39 @@
...
+import io.sphere.sdk.messages.MessageDerivateHint;
+import io.sphere.sdk.products.Product;
+import io.sphere.sdk.products.ProductProjection;
+import io.sphere.sdk.queries.PagedQueryResult;
+
+import java.time.ZonedDateTime;
+
+import static io.sphere.sdk.products.messages.MessagesPackage.*;
+
+@JsonDeserialize(as = ProductPublishedMessage.class)//important to override annotation in Message class
```
lauraluiz: Why is that?

schleichardt: The base interface has the same annotation for the default behavior of an untyped message. But the Jackson parser forces then all implementing classes to have also this annotation.

sphere-models/src/it/java/io/sphere/sdk/products/messages/ProductPublishedMessageTest.java
schleichardt: That would be a breaking change to M16, should I still rename the method to 'forMessageHint'?

lauraluiz: Let's better take a look at that on Monday to value the options we have.

lauraluiz: I know, I have a reminder in my mailbox ;)
Let's check it after lunch.

schleichardt:
@lauraluiz please review https://github.com/sphereio/sphere-jvm-sdk/commit/1fb74455fdfce257a5e908c2068c3dcb18de77c1

lauraluiz:
Good :+1:

PR #655, opened 2015-08-13
Reference<Void> or Reference<Object> should be Reference<JsonNode>. c…

schleichardt:
…lose #653 

@lauraluiz please review


lauraluiz:
It looks good, but I need some background first. Why did we have Object initially?
You can merge it and explain me later, though :)

schleichardt:
Some objects mix different types of references like this: http://dev.sphere.io/http-api-projects-productDiscounts.html#product-discount

The `Reference<Void>` or `Reference<Object>` was just a naive implementation. Now, I got a better idea to use JsonNode, so the data is better accessable and can be transformed into a concrete one like `Reference<Category>`. But the message endpoint is an exception. I left `Reference<?>`, so subclasses can override the getter with a more concrete type like `Reference<Category>`.

PR #660, opened 2015-08-14
Feature/658/taxcategory updatecommand

schleichardt:
@lauraluiz please review

### DIFF
--- a/sphere-models/src/it/java/io/sphere/sdk/taxcategories/TaxCategoryFixtures.java
+++ b/sphere-models/src/it/java/io/sphere/sdk/taxcategories/TaxCategoryFixtures.java
@@ -11,6 +11,7 @@ import io.sphere.sdk.taxcategories.queries.TaxCategoryQueryModel;
 
 import java.util.List;
 import java.util.function.Consumer;
+import java.util.function.UnaryOperator;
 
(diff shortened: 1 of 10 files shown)


sphere-models/src/it/java/io/sphere/sdk/taxcategories/TaxCategoryFixtures.java
```diff
@@ -57,4 +58,13 @@ public static void withTaxCategory(final TestClient client, final TaxCategoryDra
         user.accept(taxCategory);
         client.execute(TaxCategoryDeleteCommand.of(taxCategory));
     }
+
+    public static void withUpdateableTaxCategory(final TestClient client, final UnaryOperator<TaxCategory> user) {
```
lauraluiz: Is `user` a good parameter name for that function? Isn't it more something like `test`?

schleichardt: test is a bad name since the output is not a boolean but a TaxCategory

schleichardt:
@lauraluiz Is it now mergeable?

lauraluiz:
Yes :+1:

PR #671, opened 2015-08-19
Add methods to compare Address|TaxRate|Price without ids. close #91

### LINKED ISSUE #91: Allow comparing addresses
It is a very common practice to compare addresses (e.g. to decide if the customer selected the option "use same shipping address as billing address"). Currently the equals method have into account the ID of the address, which will always be different. 

----

* [ ] equals includes id
* [ ] alternative method exclusive id

schleichardt:
@lauraluiz please review


lauraluiz:
:+1:

PR #680, opened 2015-08-21
Remove type facet result

lauraluiz:
@schleichardt please review


sphere-common/src/main/java/io/sphere/sdk/search/SearchModelExpression.java
```diff
@@ -21,35 +21,43 @@ protected SearchModelExpression(final SearchModel<T> searchModel, final TypeSeri
         this.alias = alias;
     }
 
-    protected final String buildResultPath() {
-        return Optional.ofNullable(alias).orElse(serializedPath());
+    public final String valueSerializer(final V value) {
```
schleichardt: maybe better: serializeValue

PR #682, opened 2015-08-24
Retry quadratically

lauraluiz:
@schleichardt please review. It is just an improvement of the retry method, which now increases the delay quadratically until it reaches the maximum retry interval specified. This way we might save some time when it fails due to unsynchronized resources.

I then allowed up to 10 seconds of retry. I may lower this if I see it is not necessary, but actually if it is really unnecessary it should never have to wait that much :)


schleichardt:
:+1: try it out

PR #684, opened 2015-08-24
Untyped search model

lauraluiz:
@schleichardt please review. I've created an UntypedSearchModel that accepts simple Strings, therefore removing all type-safety that can be annoying when dealing with input from e.g. query strings.

### DIFF
--- a/sphere-common/src/main/java/io/sphere/sdk/search/RangeTermModel.java
+++ b/sphere-common/src/main/java/io/sphere/sdk/search/RangeTermModel.java
@@ -13,4 +13,19 @@ public interface RangeTermModel<T, V extends Comparable<? super V>> {
      * @return the model to specify the range of values from which to obtain the facets
      */
     RangedFacetSearchModel<T, V> faceted();
+
+    /**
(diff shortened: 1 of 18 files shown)


sphere-common/src/main/java/io/sphere/sdk/search/RangeTermModel.java
```diff
@@ -13,4 +13,19 @@
      * @return the model to specify the range of values from which to obtain the facets
      */
     RangedFacetSearchModel<T, V> faceted();
+
+    /**
+     * Gets an untyped search model (expecting simple Strings) while keeping the same search model path as this.
+     * This untyped search model allows you to build filters and facets, both range and term models.
+     * @return the untyped search model for this instance
+     */
+    default UntypedSearchModel<T> untyped() {
```
lauraluiz: If the `default` is not of your liking I could create 2 abstract classes that implement it (one for terms and one for ranges).

lauraluiz:
Now it certainly looks much better @schleichardt :D

schleichardt:
Can the branch be deleted?

PR #689, opened 2015-08-27
ProductProjection is a Versioned<Product>. close #677

### LINKED ISSUE #677: ProductProjection should be Versioned<Product>

schleichardt:
@lauraluiz please review

I am not even sure if we should do it.
But is has some charm to use `ProductProjection` as parameter for `Product` commands.


lauraluiz:
Yes, it is necessary without workarounds, since both Product and ProductProjection identify the same resource, so it is not correct that only one of classes actually identify it.
You can merge it :+1:

lauraluiz:
But take a look at the errors in Travis

PR #702, opened 2015-09-10
activate discount

yanns:
@schleichardt the test `io.sphere.sdk.products.queries.ProductQueryTest.queryProductsWithAnyDiscount` cannot success with an inactive discount

### DIFF
--- a/sphere-models/src/it/java/io/sphere/sdk/productdiscounts/ProductDiscountFixtures.java
+++ b/sphere-models/src/it/java/io/sphere/sdk/productdiscounts/ProductDiscountFixtures.java
@@ -24,7 +24,7 @@ public class ProductDiscountFixtures {
         final AbsoluteProductDiscountValue discountValue = AbsoluteProductDiscountValue.of(EURO_1);
         final LocalizedString name = en("demo product discount");
         final LocalizedString description = en("description");
-        final boolean active = false;
+        final boolean active = true;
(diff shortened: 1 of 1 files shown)


yanns:
Oups it seems I broke other tests with this change.

yanns:
The failing tests have to do with the new deprecation from @agourlay

schleichardt:
@agourlay can you give more information for the deprecation, please.

@yanns in team city tests should not fail due to deprecation warnings.

schleichardt:
Thanks Yann, I think I just should not reuse data for performance reasons if the tests behave that weird.

agourlay:
We introduced a HTTP param, on the facet endpoint called formatBooleanFacet.

If the param is provided then the booleans are formated properly i.e.```T ```becomes ```true``` and ```F``` becomes ```false```, otherwise we issue a deprecation notice.

In the far future, when all clients will have migrated to using the extra formating param, we can make it the default behavior and remove the param.

schleichardt:
So can I just add `formatBooleanFacet=true` to every search request? Or at least the ones with facets?

schleichardt:
@lauraluiz This might also interest you.

agourlay:
It is possible although the optimal solution would be to add it only to the search facets query targetting boolean attribute.
If it is sent it for any other request it will be ignored anyway.

Don't forget to change as well the way boolean facets results are deserialized.

PR #782, opened 2015-10-09
Try to fail build if command fails.

schleichardt:


hajoeichler:
@schleichardt You may simply use `set -e` at the beginning of the script.

schleichardt:
Thanks.

https://github.com/sphereio/sphere-jvm-sdk/pull/783

PR #783, opened 2015-10-09
Fix build does not fail if tests fail. close #781

### LINKED ISSUE #781: build does not fail if tests fail

schleichardt:


hajoeichler:
:+1:

PR #808, opened 2015-10-31
Fix missing Content-Length in AHC. close #799

### LINKED ISSUE #799: Missing Content-Length in Ning
Ning expects the Content-Length to be set via header parameters, while Apache calculates it itself. This makes it impossible to implement a code independent of the HTTP client while at the same time requiring the Content-Length (as you either won't have it in Ning, or Apache will have a duplicated Content-Length).

In order to fix this, we must provide the content length according to the `HttpRequestBody` type in the Ning adapter.

schleichardt:
@lauraluiz please review

@panshul007 Can you please try out a snapshot if it works for you? The version number is "1.0.0-M20-issue-799-SNAPSHOT"

You may need to add the sonatype snapshots resolver:

```
resolvers ++= Seq (
    Resolver.sonatypeRepo("releases"),
    Resolver.sonatypeRepo("snapshots"),
    Resolver.bintrayRepo("commercetools", "maven")
  )
```

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -50,7 +50,8 @@ object Build extends Build {
     .settings(libraryDependencies ++= allTestLibs.map(_ % "test,it"))
 
   lazy val `sphere-java-client-internal-test` = project.dependsOn(`sphere-java-client-core`).settings(commonSettings:_*).settings(
-    libraryDependencies ++= allTestLibs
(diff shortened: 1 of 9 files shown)


lauraluiz:
:+1: on my side.

panshul007:
:+1:

PR #813, opened 2015-11-01
improve cent amount calculation, close #789

### LINKED ISSUE #789: Add centAmount helper for MoneyImpl
* name it sphere cent amount, because it uses just 2 digits after the decimal separator
* http://mavenbrowse.pauldoo.com/central/javax/money/money-api/0.8/money-api-0.8-javadoc.jar/-/javax/money/CurrencyUnit.html#getDefaultFractionDigits%28%29

```
monetaryAmount.getNumber().numberValue(BigDecimal.class).multiply(
new BigDecimal(100)//TODO maybe 100 is not true for all currencies
).setScale(0, RoundingMode.HALF_EVEN).longValue()
```

* [x] use this in MoneyRepresentation for JsonUtil
* [x] improve money formatting documentation
* [x] improve money of cents method
* [x] tests with yen

schleichardt:
@lauraluiz please review

@cneijenhuis I also would like to get your feedback for this pull request.

### DIFF
--- a/sphere-common/src/main/java/io/sphere/sdk/json/MoneyDeserializer.java
+++ b/sphere-common/src/main/java/io/sphere/sdk/json/MoneyDeserializer.java
@@ -7,7 +7,6 @@ import io.sphere.sdk.utils.MoneyImpl;
 
 import javax.money.MonetaryAmount;
 import java.io.IOException;
-import java.math.BigDecimal;
 
 final class MoneyDeserializer extends StdScalarDeserializer<MonetaryAmount> {
     private static final long serialVersionUID = 0L;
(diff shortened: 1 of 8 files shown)


schleichardt:
Hi @sebbulon we need your expertise in a meeting to discuss this.

schleichardt:
Reason: not approved in the review since the merchant center behaves differently.

sebbulon:
ok I clicked around in the MC and observed the following things:

1. it applies commercial rounding if you enter for instance 50,551 - it will save 50,55, and if you enter 55,556 it will save 55.56 - I think this is not correct, it should validate instead and not save anything if format is not correct
2. the MC does not observe correct formatting of Yen - it actually accepts 2 fractional digits 

so it looks like there are issues in MC.
Notwidthstanding I don't quite understand right now what the issue is here- so lets discuss together on Wednesday ideally.

schleichardt:
Hmm, the JVM SDK also rounds if there are more precision than fraction digits.

The issue is, when I enter in the MC 123 Yen then it stores in the JSON 12300 as cent amount. This does not match the suggested implementation for the JVM SDK which stores 123 cent amount for 123 Yen.

see also https://support.stripe.com/questions/which-zero-decimal-currencies-does-stripe-support

sebbulon:
ok it's a bug in MC legacy -lets not fix it for now but raise it for MC2

cneijenhuis:
Nice, LGTM :+1: 

I guess the bug in MC legacy shows how easy it is to do this stuff wrong ;)

lauraluiz:
OK, let's merge it.

PR #836, opened 2015-11-22
Features/763/custom fields prices

schleichardt:


schleichardt:
@lauraluiz please review, should we rename Price into ProductPrice?

PR #877, opened 2015-12-10
Feature/875/blocking client, close #875

### LINKED ISSUE #875: blocking client
Users for synchronous scenarios used the async client quite wrong. They struggled with checked exceptions and did not incorporate explicit timeouts. For this scenarios it would be nice to have a blocking client and ease the pain.

schleichardt:


schleichardt:
@lauraluiz can you please review (even after merge)

sphere-java-client-core/src/test/java/io/sphere/sdk/client/BlockingSphereClientTest.java
```diff
@@ -0,0 +1,61 @@
...
+    @Test
+    public void timeoutPerCall() {
+        final BlockingSphereClient blockingSphereClient = newBlockingClientWhichwouldBlockForever();
+
+        final Throwable throwable = catchThrowable(() -> blockingSphereClient.executeBlocking(DummySphereRequest.of(), 1, TimeUnit.NANOSECONDS));
+
+        assertThat(throwable).isInstanceOf(SphereTimeoutException.class);
+    }
+
+    private BlockingSphereClient newBlockingClientWhichwouldBlockForever() {
```
lauraluiz: Typo: `WhichWouldBlock...` (camel case)

lauraluiz:
:+1:

sphere-java-client-core/src/main/java/io/sphere/sdk/client/BlockingSphereClientImpl.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    @Override
+    public <T> CompletionStage<T> execute(final SphereRequest<T> sphereRequest) {
+        return delegate.execute(sphereRequest);
+    }
+
+    @Override
+    public <T> T executeBlocking(final SphereRequest<T> sphereRequest) {
+        return executeBlocking(sphereRequest, defaultTimeout, unit);
```
yanns: a pitfall with that approach is that the timeout on the future can throw an exception before the timeout of the http client.
It means:
- the user receives a timeout
- but the request eventually succeed.

We have to be careful here to use a timeout long enough (more than the request timeout of the underlying http client)

schleichardt: Yes, I added a note on the construction of the timeout client.

PR #879, opened 2015-12-10
Is matching variant flag, closes #830

### LINKED ISSUE #830: support matching variants
The Search now calculates matching variants in form of an additional flag for the variant returned by the Search.
Support needs to be added to the SDK.
This is a good opportunity to add convenience method to get the matching variants I suppose, refactor existing code that duplicates the logic.

lauraluiz:
@schleichardt please review.

@agourlay also review. Check that I created some IT that currently are disabled because the flag is not working as it would be desirable. We can use them as a guide of what's left to do :)

### DIFF
--- a/sphere-models/src/it/java/io/sphere/sdk/products/search/ProductProjectionSearchMainIntegrationTest.java
+++ b/sphere-models/src/it/java/io/sphere/sdk/products/search/ProductProjectionSearchMainIntegrationTest.java
@@ -2,19 +2,31 @@ package io.sphere.sdk.products.search;
 
 import io.sphere.sdk.models.Identifiable;
 import io.sphere.sdk.products.ProductProjection;
+import io.sphere.sdk.products.ProductVariant;
+import io.sphere.sdk.products.attributes.AttributeAccess;
+import io.sphere.sdk.products.attributes.NamedAttributeAccess;
+import io.sphere.sdk.search.FilterExpression;
 import io.sphere.sdk.search.PagedSearchResult;
+import org.assertj.core.api.Condition;
+import org.junit.Ignore;
 import org.junit.Test;
 
 import java.util.List;
+import java.util.function.Function;
 import java.util.function.Predicate;
 
 import static io.sphere.sdk.test.SphereTestUtils.ENGLISH;
 import static java.lang.Math.max;
 import static java.lang.Math.min;
 import static java.util.Arrays.asList;
+import static java.util.stream.Collectors.toList;
 import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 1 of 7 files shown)


sphere-models/src/it/java/io/sphere/sdk/products/search/ProductProjectionSearchMainIntegrationTest.java
```diff
@@ -64,6 +76,79 @@ public void fuzzySearch() {
...
+
+    @Ignore // HTTP API still not having sort associated to matching variant flag
+    @Test
+    public void findMatchingVariantBySort() throws Exception {
+        final ProductProjectionSearch search = ProductProjectionSearch.ofStaged()
+                .withSort(product -> product.allVariants().attribute().ofNumber(ATTR_NAME_SIZE).byAscWithMax());
+        assertThat(executeSearch(search).getResults()).are(productWithMatchingVariantsHavingMaxSize());
+    }
+
+    @Ignore // HTTP API doesn't mark the correct variant as true when the searched info is inside the variant
```
schleichardt: Can you open a GitHub issue to memorize to reenable the tests and link the issue here?

sphere-models/src/test/resources/ProductProjectionTest/product-projection-from-search.json
```diff
@@ -0,0 +1,128 @@
...
+      },
+      {
+        "name": "longDescription",
+        "value": {
+          "en": "Some long description for the book",
+          "de": "Some long description for the book in german"
+        }
+      }
+    ],
+    "isMatchingVariant": false
```
schleichardt: @agourlay the api is not super consistent, some boolean flags start with "is" or not. 
I think it is better it leave the "is" for fields out since starting with verbs is a thing for methods.

schleichardt:
:+1:

sphere-models/src/it/java/io/sphere/sdk/products/search/ProductProjectionSearchMainIntegrationTest.java
lauraluiz: https://github.com/sphereio/sphere-jvm-sdk/issues/880

sphere-models/src/test/resources/ProductProjectionTest/product-projection-from-search.json
agourlay: I don't have strong opinion about the naming of this attribute, unfortunately it would be a breaking change now.

We can't really know who is using it so we would have to go through a deprecation process.
@mmoelli WDYT?

PR #885, opened 2015-12-11
attributes/field types renaming, close #849, close #493

### LINKED ISSUE #493: document all concrete implementing classes of AttributeType

### LINKED ISSUE #849: Merge AttributeType/FieldType
Is it possible to reuse the same AttributeTypes for ProductTypes and CustomTypes? They have exactly the same name and the similar meaning, and they are giving conflicts when using them together. It is true that is not very likely to be used together in the same class except for simple projects, but therefore it makes it a bit difficult to create a demo project.

schleichardt:
@lauraluiz please review


lauraluiz:
Great, now we know who is who :)

PR #888, opened 2015-12-11
Request gzipped responses from sphere. close #873

### LINKED ISSUE #873: Request gzipped responses from sphere

schleichardt:
@lauraluiz please review


lauraluiz:
How does it work the gzip encoding in whatever-generic-name-has-ning-now? Is it already working for it?

lauraluiz:
Is it possible to write some test for that?

schleichardt:
It is tested through the existing integration tests - for example the Apache thing failed. AHC does unpack automatically.

But I will add a test that checks the response headers so they contain gzip.

lauraluiz:
:+1:

PR #899, opened 2015-12-14
Add getCurrency() to cart and order. close #870

### LINKED ISSUE #870: should we add a helper method getCurrency() to cart/order?
which uses the total price as data source

@lauraluiz what do you think?

schleichardt:
@lauraluiz Is it ok?


lauraluiz:
Yes. TotalPrice is always there and currency should be good.

PR #900, opened 2015-12-14
Blocking wait as function

schleichardt:
@lauraluiz what do you think? @cneijenhuis need this for the payment scenario where 3 requests are performed in parallel but then a blocking wait should take place.


cneijenhuis:
Nice, thanks :+1:

sphere-common/src/main/java/io/sphere/sdk/client/SphereClientUtils.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Optional;
+import java.util.concurrent.*;
+
+public final class SphereClientUtils {
+    private SphereClientUtils() {
+    }
+
+    /**
+     * Waits with a timeout for RESPONSES of a commercetools client wrapped in a {@link CompletionStage}.
+     * This method should not ne used for other {@link CompletionStage}s since it is throwing {@link io.sphere.sdk.models.SphereException}s.
```
lauraluiz: Small typo: `should not Ne used`

src/main/java/io/sphere/sdk/meta/AsyncDocumentation.java
```diff
@@ -29,6 +29,20 @@
...
+
+ A lot of frameworks do not support asynchronous results but it is still possible to execute calls in parallel and then wait for computed result:
+
+ {@include.example io.sphere.sdk.meta.AsyncDocumentationTest#firstAsyncThenSync()}
+
+ <!--
+
+ thenCompose/thenCombine examples und dann blocking wait
+
+ -->
```
lauraluiz: Should the comment be removed?

lauraluiz:
Looks good. In the end is just moving out the blocking logic to a utils class and keeping the rest the same. I think it's good people use the proper blocking code.

PR #901, opened 2015-12-14
remove deleted review endpoint

yanns:
the review endpoint was deleted, and it will be open again in a few days/weeks.

When the review endpoint is available again, we can revert this change and adapt to the new endpoint.


schleichardt:
When the travis build is green: :+1:

PR #909, opened 2015-12-16
Ease to initialise a client with an access token. close #749

### LINKED ISSUE #749: add example/convenience method to just provide a token to initialize s sphere client
```java
SphereClientFactory.createClient(String projectKey, String token)
```

schleichardt:
@lauraluiz please review

### DIFF
--- a/sphere-java-client-core/src/main/java/io/sphere/sdk/client/SphereClientFactory.java
+++ b/sphere-java-client-core/src/main/java/io/sphere/sdk/client/SphereClientFactory.java
@@ -1,12 +1,10 @@
 package io.sphere.sdk.client;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
 import io.sphere.sdk.http.HttpClient;
+import io.sphere.sdk.http.HttpRequest;
 import io.sphere.sdk.http.HttpResponse;
-import io.sphere.sdk.json.SphereJsonUtils;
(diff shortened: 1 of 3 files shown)


sphere-java-client-core/src/main/java/io/sphere/sdk/client/SphereClientFactory.java
```diff
@@ -95,6 +93,44 @@ public SphereClient createClient(final String projectKey, final String clientId,
     }
 
     /**
+     * Creates a client which relies on an access token and does not refresh it, it reuses an existing http client.
+     *
+     * A possible use case is keeping the clientSecret secret and/or instantiate a client which is used for few requests for a limited time.
+     * The underlying http client does not get closed by closing the client
```
lauraluiz: I'm certain there's a good reason behind it I'm not thinking of right now, but why?

lauraluiz:
:+1:

sphere-java-client-core/src/main/java/io/sphere/sdk/client/SphereClientFactory.java
schleichardt: If you reuse a HTTP client for multiple sphere projects you don't want to close the HTTP client if you don't need a project anymore since then all other project requests would fail.

PR #924, opened 2016-01-08
drafts from json, close #785

### LINKED ISSUE #785: Enable Draft creation from JSON resources
At present it seems not to be possible to create drafts(e.g. ProductTypeDraft, ProductDraft) from JSON resource data, so you have to create own wrapper types and create the drafts based on the data of this types.

schleichardt:


sphere-common/src/main/java/io/sphere/sdk/models/LocalizedString.java
```diff
@@ -334,4 +334,8 @@ public String toString() {
             }
         };
     }
+
+    public static LocalizedString ofEnglish(final String translationForEnglish) {
```
lauraluiz: Are you going to deprecate the other one?

lauraluiz:
:+1: 
Luckily everything was nicely sorted and it was easy to review :P

sphere-common/src/main/java/io/sphere/sdk/models/LocalizedString.java
schleichardt: No, should I?

lauraluiz: Personally I'm not very fan of aliases in class methods, because it makes me wonder what's the difference and I waste time checking the source, plus it makes the method lists larger. Unless there's a really good reason to do it, of course. I feel `ofEnglish` is good enough already, but it's up to you.

schleichardt: I will deprecate the other.

PR #926, opened 2016-01-11
attribute read improvements, close #918

### LINKED ISSUE #918: create attribute value mapper helper
see https://github.com/sphereio/sphere-jvm-sdk/tree/attributes-improvements

do not make it string specific

schleichardt:


schleichardt:
@lauraluiz please review

sphere-models/src/it/java/io/sphere/sdk/products/attributes/LocalizedToStringProductAttributeConverterTest.java
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    @Test
+    public void convertTime() {
+        assertThat(converting(ATTR_NAME_TIME)).isEqualTo("22:05");
+    }
+
+    @Test
+    public void convertLocalizedStringSet() {
+        assertThat(converting(ATTR_NAME_LOC_TEXT_SET)).isEqualTo("");
```
lauraluiz: Why empty?

sphere-models/src/it/java/io/sphere/sdk/products/attributes/LocalizedToStringProductAttributeConverterTest.java
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    @Test
+    public void convertProductReferenceSet() {
+        assertThat(converting(ATTR_NAME_REF)).isNull();
+    }
+
+    @Test
+    public void convertProductReference() {
+        assertThat(converting(ATTR_NAME_REF_SET)).isNull();
```
lauraluiz: I would print the reference ID. I know it's not very usable, but with null we might have two bad scenarios: 1) NPE or 2) nothing shown. I would rather have the ID so at least people can obtain results (though they probably need to be adjusted).

sphere-models/src/it/java/io/sphere/sdk/products/attributes/LocalizedToStringProductAttributeConverterTest.java
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    @Ignore
+    @Test
+    public void convertCategoryReferenceSet() {
+    }
+
+    @Ignore
+    @Test
+    public void convertCategoryReference() {
```
lauraluiz: If we know the particular type of reference, cannot we show the name?

sphere-models/src/it/java/io/sphere/sdk/products/attributes/LocalizedToStringProductAttributeConverterTest.java
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    @Ignore
+    @Test
+    public void convertDoubleSet() {
+    }
+
+    @Ignore
+    @Test
+    public void convertDouble() {
```
lauraluiz: Why ignored? I guess it has something to do with int/double.

sphere-models/src/it/java/io/sphere/sdk/products/attributes/LocalizedToStringProductAttributeConverterTest.java
```diff
@@ -0,0 +1,222 @@
...
+        assertThat(converting(ATTR_NAME_DATE_TIME)).isEqualTo("11.09.01 22:05");
+    }
+
+    @Test
+    public void convertDate() {
+        assertThat(converting(ATTR_NAME_DATE)).isEqualTo("11.09.2001");
+    }
+
+    @Test
+    public void convertBoolean() {
```
lauraluiz: Maybe I'm not finding the order, but I feel all methods are unsorted, and it makes it difficult to see if we are missing some attribute :P

sphere-models/src/it/java/io/sphere/sdk/products/search/ProductProjectionSearchModelIntegrationTest.java
```diff
@@ -1,150 +1,49 @@
...
+        product2 = data.getProduct2();
+    }
 
-        productSomeId = findBySku.apply(SKU_SOME_ID).orElseGet(() -> createTestProduct("Some Id", ProductVariantDraftBuilder.of().sku(SKU_SOME_ID).build()));
-        productOtherId = findBySku.apply(SKU_OTHER_ID).orElseGet(() -> createTestProduct("Other Id", ProductVariantDraftBuilder.of().sku(SKU_OTHER_ID).build()));
-        product1 = findBySku.apply(SKU1).orElseGet(() -> createProduct1());
-        product2 = findBySku.apply(SKU2).orElseGet(() -> createProduct2());
+    @AfterClass
+    public static void clean() {
+        destroy(client(), data);
```
lauraluiz: With this we don't have anymore fixed products in the project and the tests might fail again. I know we kind of have this effect in TC, but here it became quite annoying for the developing process.

sphere-models/src/main/java/io/sphere/sdk/products/attributes/LocalizedToStringProductAttributeConverter.java
```diff
@@ -0,0 +1,231 @@
...
+    protected String convertDate(final LocalDate dateValue, final Attribute attribute, final ProductType productType) {
+        final DateTimeFormatter dateTimeFormatter = DateTimeFormatter.ofLocalizedDate(FormatStyle.MEDIUM).withLocale(locale());
+        return dateTimeFormatter.format(dateValue);
+    }
+
+    @Override
+    protected String convertBoolean(final Boolean booleanValue, final Attribute attribute, final ProductType productType) {
+        return booleanValue.toString();
+    }
+}
```
lauraluiz: I don't quite get why you pre-sort some, others post-sort and others simply not sorted. Are the results having arbitrary order? I would expect that CTP delivers it sorted as they merchant creates them.

sphere-models/src/it/java/io/sphere/sdk/products/attributes/LocalizedToStringProductAttributeConverterTest.java
schleichardt: Good catch! I will fix that.

sphere-models/src/it/java/io/sphere/sdk/products/attributes/LocalizedToStringProductAttributeConverterTest.java
schleichardt: Please elaborate.

sphere-models/src/it/java/io/sphere/sdk/products/search/ProductProjectionSearchModelIntegrationTest.java
schleichardt: For other tests we may need to delete all products anyway.

sphere-models/src/main/java/io/sphere/sdk/products/attributes/LocalizedToStringProductAttributeConverter.java
schleichardt: I think date is a good example:
the dates: 15.10. and 20.09
sorted as Date: 20.09., 15.10.
sorted as String: 15.10., 20.09.

So for dates it is better to sort the values first and then use `toString()`.

For localized Strings it is better to sort the translated Strings rather than the objects containing all information.

sphere-models/src/it/java/io/sphere/sdk/products/attributes/LocalizedToStringProductAttributeConverterTest.java
lauraluiz: From a category, for example, as we know it's a category and the fields categories have, by default I would show the `ref.obj.name` + locale if the reference is expanded and the `ref.id` if it's not. Just a nice-to-have, of course. Mostly asking if it's possible.

sphere-models/src/it/java/io/sphere/sdk/products/search/ProductProjectionSearchModelIntegrationTest.java
lauraluiz: That's bad, because we already reached the conclusion not so long ago that with eventual consistency only keeping the products was a solution :P
Anyway if it works, then there's no problem.

sphere-models/src/main/java/io/sphere/sdk/products/attributes/LocalizedToStringProductAttributeConverter.java
lauraluiz: Okay that is clear. So the question is, do we want to sort them? Do they come with a predetermined order from the platform (the order the merchant gives them)? Or they are received in a random order?

schleichardt: That is the evil thing - they are sets!

PR #1005, opened 2016-02-25
rename and open LocalizedToStringProductAttributeConverter, close #992

### LINKED ISSUE #992: LocalizedToStringProductAttributeConverter public constructor
`LocalizedToStringProductAttributeConverter` class should have a public constructor (maybe static factory constructor) to allow to use it directly without needing to inherit from it (which is default behaviour). In case you want to customized some behaviour because it doesn't satisfy you, then you are more than welcome to create a subclass.

Also, add in the documentation where to obtain the reference for the product type (product).

And maybe I've already asked it, but why `LocalizedToString` prefix? I can get the strings are localized, but it makes the name painfully long, maybe we can skip the obvious parts of it unless it really makes sense (i.e. there's a non-localized converted).

schleichardt:
@lauraluiz please review

release notes:

```html
 <li class=change-in-release>rename {@code LocalizedToStringProductAttributeConverter} 
to {@link io.sphere.sdk.products.attributes.DefaultProductAttributeFormatter}</li>
 <li class=new-in-release>make constructors of 
{@link io.sphere.sdk.products.attributes.DefaultProductAttributeFormatter} public</li>
 <li class=new-in-release>add alias
 {@link io.sphere.sdk.products.attributes.DefaultProductAttributeFormatter#format(Attribute, Referenceable)}
 for
 {@link io.sphere.sdk.products.attributes.DefaultProductAttributeFormatter#convert(Attribute, Referenceable)}</li>
```

### DIFF
--- a/commercetools-internal-docs/src/main/java/io/sphere/sdk/meta/ReleaseNotes.java
+++ b/commercetools-internal-docs/src/main/java/io/sphere/sdk/meta/ReleaseNotes.java
@@ -28,8 +28,6 @@ import io.sphere.sdk.products.*;
 import io.sphere.sdk.products.attributes.Attribute;
 import io.sphere.sdk.products.attributes.AttributeAccess;
 import io.sphere.sdk.products.attributes.AttributeDefinition;
-import io.sphere.sdk.products.attributes.LocalizedToStringProductAttributeConverter;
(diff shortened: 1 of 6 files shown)


lauraluiz:
Shorter! :+1:

PR #1016, opened 2016-02-29
activate rules

schleichardt:


schleichardt:
see https://github.com/sphereio/sphere-jvm-sdk/pull/1018

PR #1062, opened 2016-03-15
Small changes

schleichardt:


schleichardt:
@lauraluiz please review, reviewing every single commit would be easier, I think

lauraluiz:
:+1:

PR #1136, opened 2016-05-13
improve CompletableFutureUtils, close #1123, close #1124, close #1126, close #1128

### LINKED ISSUE #1123: Document CompletableFutureUtils
The methods there are great, so it will be very useful to be able to use them and for that we need docu, as I'm going to use them in Sunrise.

### LINKED ISSUE #1124: add listOfFuturesToFutureOfList from Sunrise
* [ ] replace FutureUtils in Sunrise

```java
package io.sphere.sdk.utils;

import java.util.List;
import java.util.concurrent.CompletableFuture;
import java.util.concurrent.CompletionStage;
import java.util.stream.Collectors;

import static java.util.stream.Collectors.toList;

public final class FutureUtils {

    private FutureUtils() {
    }

    /**
     * Transforms a list of {@code CompletionStage} into a {@code CompletionStage} of a list,
     * that will be completed once all the elements of the given list are completed.
     * @param stageList list of {@code CompletionStage}
     * @param <T> the element obtained from the list of {@code CompletionStage}
     * @return the {@code CompletionStage} of a list of elements
     */
    public static <T> CompletionStage<List<T>> listOfFuturesToFutureOfList(final List<CompletionStage<T>> stageList) {
        final List<CompletableFuture<T>> futureList = stageList.stream()
                .map(CompletionStage::toCompletableFuture)
                .collect(toList());
        final CompletableFuture[] futuresAsArray = futureList.toArray(new CompletableFuture[futureList.size()]);
        return CompletableFuture.allOf(futuresAsArray)
                .thenApply(x -> futureList.stream()
                        .map(CompletableFuture::join)
                        .collect(Collectors.<T>toList()));
    }
}
```

### LINKED ISSUE #1126: fix recover with and add executor
```java
 public static <T> CompletionStage<T> recoverWith(final CompletionStage<T> future, final Executor executor, final Function<? super Throwable, CompletionStage<T>> f) {
        final CompletableFuture<T> result = new CompletableFuture<>();
        final BiConsumer<T, Throwable> action = (value, error) -> {
            if (value != null) {
                result.complete(value);
            } else {
                final CompletionStage<T> alternative = f.apply(error);
                alternative.whenComplete((alternativeValue, alternativeError) -> {
                    if (alternativeValue != null) {
                        result.complete(alternativeValue);
                    } else {
                        result.completeExceptionally(alternativeError);
                    }
                });
            }
        };
        future.whenCompleteAsync(action, executor);
        return result;
    }
```

### LINKED ISSUE #1128: create alias for failed futures
see https://github.com/sphereio/commercetools-sunrise-java/pull/363/files/1f4bc4db3e8f81358e4c97a8e32fc0cc3412bd4b..475cfda327f548e669ae09e023cfbba863257e8a#r62505961

schleichardt:
@lauraluiz please review

### DIFF
--- a/commercetools-sdk-base/src/main/java/io/sphere/sdk/utils/CompletableFutureUtils.java
+++ b/commercetools-sdk-base/src/main/java/io/sphere/sdk/utils/CompletableFutureUtils.java
@@ -1,10 +1,7 @@
 package io.sphere.sdk.utils;
 
 import java.util.NoSuchElementException;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionException;
-import java.util.concurrent.CompletionStage;
-import java.util.concurrent.ExecutionException;
+import java.util.concurrent.*;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import java.util.function.Function;
@@ -69,22 +66,26 @@ public final class CompletableFutureUtils {
     }
 
     public static <T> CompletionStage<T> recoverWith(final CompletionStage<T> future, final Function<? super Throwable, CompletionStage<T>> f) {
+        return recoverWith(future, f, ForkJoinPool.commonPool());
+    }
+
+    public static <T> CompletionStage<T> recoverWith(final CompletionStage<T> future, final Function<? super Throwable, CompletionStage<T>> f, final Executor executor) {
         final CompletableFuture<T> result = new CompletableFuture<>();
         final BiConsumer<T, Throwable> action = (value, error) -> {
             if (value != null) {
                 result.complete(value);
             } else {
                 final CompletionStage<T> alternative = f.apply(error);
-                alternative.whenComplete((alternativeValue, alternativeError) -> {
+                alternative.whenCompleteAsync((alternativeValue, alternativeError) -> {
                     if (alternativeValue != null) {
                         result.complete(alternativeValue);
                     } else {
                         result.completeExceptionally(alternativeError);
                     }
-                });
+                }, executor);
             }
         };
-        future.whenComplete(action);
+        future.whenCompleteAsync(action, executor);
         return result;
     }
 
--- a/commercetools-sdk-base/src/main/java/io/sphere/sdk/utils/CompletableFutureUtils.java
+++ b/commercetools-sdk-base/src/main/java/io/sphere/sdk/utils/CompletableFutureUtils.java
@@ -125,7 +125,7 @@ public final class CompletableFutureUtils {
     }
 
     /**
-     * Creates a {@link CompletionStage} which may recovered from an occurred error. Alias for {@link CompletionStage#exceptionally(Function)}.
+     * Creates a {@link CompletionStage} which can be recovered if an error occurs. Alias for {@link CompletionStage#exceptionally(Function)}.
      * If the recovery function also requires to create a {@link CompletionStage} then use {@link #recoverWith(CompletionStage, Function)}.
      *
      * {@include.example io.sphere.sdk.utils.CompletableFutureUtilsTest#recoverFailure()}
@@ -133,14 +133,14 @@ public final class CompletableFutureUtils {
      * @param future future which may completes exceptionally
      * @param f function how the exception should be handled to create a successfully completed future, but it also can throw exceptions
      * @param <T> type of the value of the future
-     * @return a future which may recovered from errors
+     * @return a future which can be recovered from errors
      */
     public static <T> CompletionStage<T> recover(final CompletionStage<T> future, final Function<Throwable, ? extends T> f) {
         return future.exceptionally(f);
     }
 
     /**
-     * Creates a {@link CompletionStage} which may recovered from an occurred error.
+     * Creates a {@link CompletionStage} which can be recovered if an error occurs.
      * If the recovery function does not require to create a {@link CompletionStage} then use {@link #recover(CompletionStage, Function)}.
      *
      * {@include.example io.sphere.sdk.utils.CompletableFutureUtilsTest#recoverWithSuccessInSecond()}
@@ -148,14 +148,14 @@ public final class CompletableFutureUtils {
      * @param future future which may completes exceptionally
      * @param f function how the exception should be handled to create a successfully completed future
      * @param <T> type of the value of the future
-     * @return a future which may recovered from errors
+     * @return a future which can be recovered from errors
      */
     public static <T> CompletableFuture<T> recoverWith(final CompletionStage<T> future, final Function<? super Throwable, CompletionStage<T>> f) {
         return recoverWith(future, f, ForkJoinPool.commonPool());
     }
 
     /**
-     * Creates a {@link CompletionStage} which may recovered from an occurred error by excuting a function in a certain thread pool.
+     * Creates a {@link CompletionStage} which can be recovered if an error occurs by executing a function in a certain thread pool.
      * If the recovery does not require to create a {@link CompletionStage} then use {@link #recover(CompletionStage, Function)}.
      *
      * Have a look at {@link #recoverWith(CompletionStage, Function)} for an example without the thread pool.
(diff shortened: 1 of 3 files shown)


schleichardt:
preview for the Javadoc: http://schleichardt.github.io/preview-future-utils/apidocs/io/sphere/sdk/utils/CompletableFutureUtils.html

commercetools-sdk-base/src/main/java/io/sphere/sdk/utils/CompletableFutureUtils.java
```diff
@@ -17,6 +18,15 @@
...
+    /**
+     * Creates a {@link CompletableFuture} which is completed successfully with the given object.
+     *
+     * {@include.example io.sphere.sdk.utils.CompletableFutureUtilsTest#testSuccessful()}
+     *
+     * @param object the result of the future
+     * @param <T> the type of the object
+     * @return future
+     */
     public static <T> CompletableFuture<T> successful(final T object) {
```
lauraluiz: I like the "pair" of `failed`, good idea :D

commercetools-sdk-base/src/main/java/io/sphere/sdk/utils/CompletableFutureUtils.java
```diff
@@ -64,49 +124,178 @@ private CompletableFutureUtils() {
...
+     * @param <T> type of the value of the future
+     * @return a future which may recovered from errors
+     */
     public static <T> CompletionStage<T> recover(final CompletionStage<T> future, final Function<Throwable, ? extends T> f) {
         return future.exceptionally(f);
     }
 
-    public static <T> CompletionStage<T> recoverWith(final CompletionStage<T> future, final Function<? super Throwable, CompletionStage<T>> f) {
+    /**
+     * Creates a {@link CompletionStage} which may recovered from an occurred error.
```
lauraluiz: *"which may recover from..."

commercetools-sdk-base/src/main/java/io/sphere/sdk/utils/CompletableFutureUtils.java
```diff
@@ -64,49 +124,178 @@ private CompletableFutureUtils() {
...
+     * If the recovery function does not require to create a {@link CompletionStage} then use {@link #recover(CompletionStage, Function)}.
+     *
+     * {@include.example io.sphere.sdk.utils.CompletableFutureUtilsTest#recoverWithSuccessInSecond()}
+     *
+     * @param future future which may completes exceptionally
```
lauraluiz: *complete

commercetools-sdk-base/src/main/java/io/sphere/sdk/utils/CompletableFutureUtils.java
lauraluiz: Actually I would rather say "which is recovered from an error, in case an error occurs", since the possibility is on the error occurring, not on the recoverability. Or simply "which is recovered if an error occurs".

lauraluiz: Actually, not. You can let an error happen anyway. So "which can be recovered if an error occurs". Because the developer has the power to recover it via this function. Boy, that's difficult phrasing :D

commercetools-sdk-base/src/main/java/io/sphere/sdk/utils/CompletableFutureUtils.java
```diff
@@ -64,49 +124,178 @@ private CompletableFutureUtils() {
...
+     * @param f function how the exception should be handled to create a successfully completed future
+     * @param <T> type of the value of the future
+     * @return a future which may recovered from errors
```
lauraluiz: "which can be recovered from errors"

commercetools-sdk-base/src/main/java/io/sphere/sdk/utils/CompletableFutureUtils.java
```diff
@@ -64,49 +124,178 @@ private CompletableFutureUtils() {
...
+     * If the recovery does not require to create a {@link CompletionStage} then use {@link #recover(CompletionStage, Function)}.
+     *
+     * Have a look at {@link #recoverWith(CompletionStage, Function)} for an example without the thread pool.
+     *
+     * @param future future which may completes exceptionally
+     * @param f function how the exception should be handled to create a successfully completed future
+     * @param executor thread pool to execute the recover function
+     * @param <T> type of the value of the future
+     * @return a future which may recovered from errors
+     */
```
lauraluiz: Apply fixes from last method here, if any.

commercetools-sdk-base/src/main/java/io/sphere/sdk/utils/CompletableFutureUtils.java
```diff
@@ -64,49 +124,178 @@ private CompletableFutureUtils() {
...
         };
-        future.whenComplete(action);
+        future.whenCompleteAsync(action, executor);
         return result;
     }
 
-    public static <T, X extends Throwable> T orElseThrow(final CompletionStage<T> stage, Supplier<? extends X> exceptionSupplier) throws X, ExecutionException, InterruptedException {
+    /**
+     *
+     * Tries to access the completed value of the future and returns it/or throws the exception inside the future if already available or throws the given exception.
```
lauraluiz: That's a complex sentence. Let's try:
"Tries to access the completed future if available and returns its value (or exception in case the future completed exceptionally), otherwise throws the given exception." What do you think?

commercetools-sdk-base/src/main/java/io/sphere/sdk/utils/CompletableFutureUtils.java
```diff
@@ -64,49 +124,178 @@ private CompletableFutureUtils() {
...
+            return future.join();
         } else {
             throw exceptionSupplier.get();
         }
     }
 
-    public static <T> T orElseGet(final CompletionStage<T> stage, final Supplier<T> other) throws ExecutionException, InterruptedException {
+    /**
+     *
+     * Tries to access the completed value of the future and returns it/or throws the exception inside the future if already available or uses the supplier to get a default value.
```
lauraluiz: Apply same fixes as previous method, if any.

commercetools-sdk-base/src/main/java/io/sphere/sdk/utils/CompletableFutureUtils.java
```diff
@@ -64,49 +124,178 @@ private CompletableFutureUtils() {
...
 
-    public static <T, X extends Throwable> T orElseThrow(final CompletionStage<T> stage, Supplier<? extends X> exceptionSupplier) throws X, ExecutionException, InterruptedException {
+    /**
+     *
+     * Tries to access the completed value of the future and returns it/or throws the exception inside the future if already available or throws the given exception.
+     *
+     * {@include.example io.sphere.sdk.utils.CompletableFutureUtilsTest#orElseThrow()}
+     *
+     * @param stage the future
+     * @param exceptionSupplier code which should be executed when the value is not available yet or the future completed exceptionally
```
lauraluiz: So the exception coming from the exceptionally completed future will be replaced by the exception I'm defining?

lauraluiz: Sounds like it's trying to recover, which is not the idea I get from the description.

schleichardt: It is kind of a time is up method, if it is not done, throw the exception.
If the stage completed exceptionally than the futures exception will win.
The supplied exception only wins if the future is not completed.

lauraluiz: That's what I know from the method. But the description of the parameter doesn't say that as it's written.
"code which should be executed if the future is not yet completed (either successfully or exceptionally)"

lauraluiz:
👍  Much better now, great to have all the documentation and examples! Now if people don't use it is because they cannot find the class, which is another story :D

commercetools-sdk-base/src/main/java/io/sphere/sdk/utils/CompletableFutureUtils.java
schleichardt: I just use the names from scalas Future class/companion object.

commercetools-sdk-base/src/main/java/io/sphere/sdk/utils/CompletableFutureUtils.java
schleichardt: great, easier to understand

PR #1140, opened 2016-05-17
inline ProductDataFilterSearchModel, close #1057

### LINKED ISSUE #1057: inline ProductDataFilterSearchModel
@lauraluiz why do we need that class in between?

schleichardt:
@lauraluiz please review


lauraluiz:
👍  thanks, much better

PR #1141, opened 2016-05-17
classes as interfaces, close #1053

### LINKED ISSUE #1053: check classes if they should be interfaces

schleichardt:
@lauraluiz please review


lauraluiz:
👍 
I didn't look super in detail, because I know what you did (creating the interfaces and maybe moving some methods as default methods). If there is some particularity that I should look at, let me know.

PR #1163, opened 2016-07-07
Feature/1152/new option flag for recalculate

salander:
Add new option flag for recalculate to update product data.
Add test for new recalculate flag.


schleichardt:
Excellent work that helped to speed up the process. Do you need a release tomorrow or could you also wait until Monday @salander 

I remove the new of method in case of more flags are coming and the of method would then be ambiguous.

salander:
@schleichardt no need for a release today. can wait for monday.

PR #1164, opened 2016-07-07
anonymousId, close #1162

### LINKED ISSUE #1162: add anonymousId
* [ ] add anonymousId to OrderLike
* [ ] anonymousId can be set when creating a cart.
* [ ]  The anonymousId can be provided when a customer signs up
* [ ]  The anonymousId can be provided when a customer signs in to assign carts and orders to a customer account.

http://dev.commercetools.com/release-notes.html#release-notes---commercetools-platform---version-release-27062016

schleichardt:
@cneijenhuis can you have a look at the tests if the assertions are right, please?

### DIFF
--- a/commercetools-models/src/main/java/io/sphere/sdk/customers/commands/CustomerSignInCommand.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/customers/commands/CustomerSignInCommand.java
@@ -38,11 +38,14 @@ public final class CustomerSignInCommand extends CommandImpl<CustomerSignInResul
     private final String password;
     @Nullable
     private final String anonymousCartId;
+    @Nullable
+    private final String anonymousId;
 
-    private CustomerSignInCommand(final String email, final String password, @Nullable final String anonymousCartId) {
+    private CustomerSignInCommand(final String email, final String password, @Nullable final String anonymousCartId, @Nullable final String anonymousId) {
         this.email = email;
         this.password = password;
         this.anonymousCartId = anonymousCartId;
+        this.anonymousId = anonymousId;
     }
 
     public static CustomerSignInCommand of(final String email, final String password) {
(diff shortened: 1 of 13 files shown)


commercetools-models/src/main/java/io/sphere/sdk/customers/commands/CustomerSignInCommand.java
```diff
@@ -38,19 +38,26 @@
...
         return of(email, password, null);
     }
 
     public static CustomerSignInCommand of(final String email, final String password, @Nullable final String anonymousCartId) {
-        return new CustomerSignInCommand(email, password, anonymousCartId);
+        return new CustomerSignInCommand(email, password, anonymousCartId, null);
+    }
+
+    public CustomerSignInCommand withAnonymousId(@Nullable final String anonymousId) {
+        return new CustomerSignInCommand(email, password, anonymousCartId, anonymousId);
```
cneijenhuis: Out of curiosity: Why is the `anonymousCartId` in the `of`, but not the `anonymousId`?

cneijenhuis:
Tests LGTM 👍 .

The beauty is not shown in the tests, however 😉 If you create a cart, order it, and then create a new active cart, all three will be added to the customer.

commercetools-models/src/main/java/io/sphere/sdk/customers/commands/CustomerSignInCommand.java
schleichardt: historical grown technical dept, I did not assume that the number of possible parameters will grow and ofEmailPasswordAnonymousId sounded weird. With withers we can easily grow.

I don't like `public static CustomerSignInCommand of(final String email, final String password, @Nullable final String anonymousCartId)` anymore. I would just use the "of" with two parameters and also use for anonymousCartId a wither.

cneijenhuis: got it 👍

butenkor:
@cneijenhuis Regarding yor last statement  "create a new active cart, all three will be added to the customer". How new active cart is linked to the customer if i do not set anonymousid to new cart first?

butenkor:
Or i need to set customer on cart creation?

cneijenhuis:
@butenkor Depends, if you use the `/me` endpoints (which are available in PHP and iOS SDKs so far), it will automagically! 🎉 
If you use the regular endpoints, as shown here, you need to set the anonymousId on the new cart.

For more info I have written a tutorial: http://dev.commercetools.com/tutorial-anonymous-session.html

Issue #1156, opened 2016-06-10
add ProductProjectionSearch.ofCurrent().bySlug(locale, slug)


schleichardt:

mht-dotsource:
@schleichardt the other methods follow the naming "with..." or "plus..." e.g. withPriceSelection() should i follow the naming convention or is the bySlug somehow different?

schleichardt:
it is just to this io.sphere.sdk.products.queries.ProductProjectionQuery#bySlug for the search endpoint.

I wouldn't mind if you also provide bySku as in query.

PR #1167, opened 2016-07-08
added bySlug and bySku with integrationtest #1156

mht-dotsource:
@schleichardt pls review 

https://github.com/commercetools/commercetools-jvm-sdk/blob/master/CONTRIBUTING.md is leading into 404. Is there any contributing guidelines I should follow.

I wrote Integrationtests that works fine. I didn't know if I should also write some Unit-Tests beacuse the code just calls already unit-tested methods


schleichardt:
great, thanks

PR #1212, opened 2016-09-09
fix JsonNodeQuery does not accept URL query parameters, close #1208

### LINKED ISSUE #1208: JsonNodeQuery does not accept URL query parameters
* for example for product projections staged=true

schleichardt:
@yanns please review


schleichardt:
solves #1208

yanns:
👍

PR #1218, opened 2016-09-21
Fix tests

peter-gerhard:
@yanns fixes the tests for set attribute due to breaking changes
https://sphere.atlassian.net/wiki/display/S/Breaking+Changes+between+Facet+and+Aggregation+Framework


peter-gerhard:
Checks should fail on ci now, since the env variable is not set to expect aggregations but ci is configured to use them.

yanns:
👍

PR #1219, opened 2016-09-22
Feature/1120/facet strategy

peter-gerhard:
@yanns This PR enables to use product counts on facets.
I refrained to put more integration test to not add more conditional assertions.
pls review.

### DIFF
--- a/commercetools-models/src/main/java/io/sphere/sdk/products/search/ProductCategoriesIdTermFacetSearchModel.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/products/search/ProductCategoriesIdTermFacetSearchModel.java
@@ -25,7 +25,7 @@ public interface ProductCategoriesIdTermFacetSearchModel<T> extends TermFacetSea
     ProductCategoriesIdTermFacetSearchModel<T> withAlias(final String alias);
 
     @Override
-    ProductCategoriesIdTermFacetSearchModel<T> countingProducts(final Boolean isCountingProducts);
+    ProductCategoriesIdTermFacetSearchModel<T> withCountingProducts(final Boolean isCountingProducts);
 
     FilteredFacetExpression<T> onlyTermSubtree(final String categoryId);
 
--- a/commercetools-models/src/main/java/io/sphere/sdk/products/search/ProductCategoriesIdTermFacetSearchModel.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/products/search/ProductCategoriesIdTermFacetSearchModel.java
@@ -24,6 +24,9 @@ public interface ProductCategoriesIdTermFacetSearchModel<T> extends TermFacetSea
     @Override
     ProductCategoriesIdTermFacetSearchModel<T> withAlias(final String alias);
 
+    @Override
+    ProductCategoriesIdTermFacetSearchModel<T> countingProducts(final Boolean isCountingProducts);
+
     FilteredFacetExpression<T> onlyTermSubtree(final String categoryId);
 
     /**
--- a/commercetools-sdk-base/src/main/java/io/sphere/sdk/search/model/FilteredFacetExpressionImpl.java
+++ b/commercetools-sdk-base/src/main/java/io/sphere/sdk/search/model/FilteredFacetExpressionImpl.java
@@ -26,8 +26,4 @@ class FilteredFacetExpressionImpl<T, V> extends TermExpression<T, V> implements 
         return super.alias();
     }
 
-    @Override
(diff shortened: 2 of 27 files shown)


commercetools-sdk-base/src/main/java/io/sphere/sdk/search/model/FilteredFacetExpressionImpl.java
```diff
@@ -25,4 +25,9 @@ public boolean equals(Object o) {
     public String alias() {
         return super.alias();
     }
+
+    @Override
```
peter-gerhard: I followed the alias example here. Maybe  im missing something but what is the sense of overriding a method to call `super.method` ?

schleichardt:
I would give the review to @lauraluiz since she is the expert for the facets

commercetools-models/src/main/java/io/sphere/sdk/products/search/ProductCategoriesIdTermFacetSearchModel.java
```diff
@@ -24,6 +24,9 @@
     @Override
     ProductCategoriesIdTermFacetSearchModel<T> withAlias(final String alias);
 
+    @Override
+    ProductCategoriesIdTermFacetSearchModel<T> countingProducts(final Boolean isCountingProducts);
```
lauraluiz: Following the structure of `alias` I would call it `withCountingProducts`. It is consistent with the semantics throughout the SDK, as it is returning an object with the same characteristics as the current one but with the counting products to whatever value you gave. The other methods are moving forward instead.

commercetools-sdk-base/src/main/java/io/sphere/sdk/search/model/FilteredFacetExpressionImpl.java
lauraluiz: Probably the case is to enable the highlighting of the method in the developer's IDE. Otherwise the method (which should be an important one) is relegated to a less important role and developers might have trouble finding it. If that's not the case, then it's a mistake :) You can try it out in the tests.

schleichardt: if it is for the IDE highlighting it should not be in the impl class but on the class you use

commercetools-sdk-base/src/test/java/io/sphere/sdk/search/model/FacetExpressionTest.java
```diff
@@ -62,6 +62,24 @@ public void facetsByLessThanRange() throws Exception {
                 assertThat(expr).isEqualTo("path.to.attribute:range(* to 3)"));
     }
 
+    @Test
+    public void facetsWithAlias() throws Exception {
```
lauraluiz: 👍

commercetools-sdk-base/src/main/java/io/sphere/sdk/search/model/FilteredFacetExpressionImpl.java
lauraluiz: Yep, that's precisely the reason why I left open the option of the mistake, because it's an impl class and that doesn't sound correct. But maybe it's a special case, I can't remember. That's why better let's test it.

commercetools-models/src/main/java/io/sphere/sdk/products/search/ProductCategoriesIdTermFacetSearchModel.java
schleichardt: @lauraluiz shouldn't it be `withCountingProducts(param)` and `Boolean isCountingProducts()` ? The `is` is part of the bean getter.

peter-gerhard: I changed it to `withCountingProducts` now since both of you suggested so

lauraluiz: I was doubting how we did it in the past, I want to be consistent. But I couldn't remember any situation. In any case, I think it reads better *without* the `is`.

peter-gerhard: Thats true but the property of the class is called `isSomething` of `hasSomething` for Boolean properties. Dropping this breaks the symmetry `with<property-name>`. 
Its a matter of taste.

PR #1222, opened 2016-09-28
Product updates via sku improvements

schleichardt:
@yanns @ccsalazarr-commercetools please review

I changed some signatures to reduce confusion. So if the name of the method starts with `ofSku` then the first parameter should be the SKU, especially `public static RemoveImage ofSku(final String imageUrl, final String sku)` was counterintuitive.

In the docu I also removed the classic `of` method links since it is better to use only the named `of` methods like `ofSku`.

### DIFF
--- a/commercetools-models/src/main/java/io/sphere/sdk/products/commands/updateactions/AddExternalImage.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/products/commands/updateactions/AddExternalImage.java
@@ -12,8 +12,6 @@ import javax.annotation.Nullable;
  *
  *  {@doc.gen intro}
  *
- *  {@include.example io.sphere.sdk.products.commands.ProductUpdateCommandIntegrationTest#addExternalImage()}
- *
  *  <p>By variant ID (every variant has a variantId):</p>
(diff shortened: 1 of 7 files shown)


commercetools-models/src/main/java/io/sphere/sdk/products/commands/updateactions/AddExternalImage.java
```diff
@@ -41,11 +39,11 @@ public static AddExternalImage of(final Image image, final Integer variantId) {
         return new AddExternalImage(image, variantId, null);
     }
 
-    public static AddExternalImage ofVariantId(final Image image, final Integer variantId) {
+    public static AddExternalImage ofVariantId(final Integer variantId, final Image image) {
```
yanns: this is a new method right? We do not break backwards compatibility here?

schleichardt: yes, changing the signature of `ofVariantId` and `ofSku` is safe since they were not included in the previous release

yanns: 👍

ccsalazarr-commercetools:
👍

PR #1224, opened 2016-09-28
Set shipping method of Id, close #920

### LINKED ISSUE #920: SetShippingMethod.ofId
`SetShippingMethod.ofId(shippingMethodId)`
to cover the use case
`SetShippingMethod.of(Reference.of(ShippingMethod.referenceTypeId(), shippingMethodId))`

ccsalazarr-commercetools:


commercetools-models/src/main/java/io/sphere/sdk/carts/commands/updateactions/SetShippingMethod.java
```diff
@@ -35,6 +35,12 @@ public static SetShippingMethod of(@Nullable final Referenceable<ShippingMethod>
                 : ofRemove();
     }
 
+    public static SetShippingMethod ofId(@Nullable String shippingMethodId){
```
schleichardt: a space character is missing in `){`

PR #1226, opened 2016-09-29
Missing state update actions, close #1013

### LINKED ISSUE #1013: add missing state update actions
http://dev.commercetools.com/http-api-projects-states.html

ccsalazarr-commercetools:


commercetools-models/src/test/java/io/sphere/sdk/states/commands/StateUpdateCommandIntegrationTest.java
```diff
@@ -92,4 +95,39 @@ public void changeType() {
...
+            final Set<StateRole> roles = asSet(REVIEW_INCLUDED_IN_STATISTICS);
+            final StateUpdateCommand command = StateUpdateCommand.of(state, SetRoles.of(roles));
+            final State updatedState = client().executeBlocking(command);
+            assertThat(updatedState.getRoles()).isEqualTo(roles);
+            return updatedState;
+        });
+    }
+
+    @Test
+    public void addRoles() throws Exception {
```
schleichardt: In this test I would have preferred that `AddRoles` is used with a state that has already roles otherwise `SetRoles` could have been used.

PR #1229, opened 2016-09-29
add CartShippingInfo.getDiscountedPrice and expansion path dsl, close #1225

### LINKED ISSUE #1225: Unable to access CartShippingInfo.getDiscountedPrice?
I have a shipping promo with a predicate that applies a 100% discount to the shippingCost when customerGroup and shippingMethodName match.  This all seems to work fine.

When applied, the cart.totalPrice is correct, and it excludes the defined Price for the shippingMethod -- as it must take in to account the cartShippingInfo.getDiscountedPrice.

In the impex playground, if I expand shippingInfo.discountedPrice.includedDiscounts[*].discount in a cart query, it returns the full object in the response, but none of that seems to make it in to the Java Objects when deserialized.

What I'm left with currently is a bit of post-processing to determine if free shipping discount was applied, and display either a discount lineItem in the summary, or just overwriting the shipping price by hand so everything in the order summary adds up correctly.

Is there an easy way that I'm not seeing to get at this info with the current sdk release?

[cartQueryWithShippingDiscount.txt](https://github.com/commercetools/commercetools-jvm-sdk/files/499117/cartQueryWithShippingDiscount.txt)

schleichardt:
@yanns please review, even I merged it to be included into the release

### DIFF
--- a/commercetools-models/src/main/java/io/sphere/sdk/carts/CartShippingInfo.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/carts/CartShippingInfo.java
@@ -1,6 +1,7 @@
 package io.sphere.sdk.carts;
 
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import io.sphere.sdk.cartdiscounts.DiscountedLineItemPrice;
 import io.sphere.sdk.models.Reference;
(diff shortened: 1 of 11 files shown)


yanns:
👍

Issue #32, opened 2014-07-11
Add helper to generate unique orderNumbers/customerNumbers


hajoeichler:
In order to avoid duplicates in order numbers (or customer numbers), it would be nice to get a helper that generates a project-wide unique number.

schleichardt:
see also https://github.com/sphereio/dev.commercetools.com/issues/61

@lauraluiz You write the implementation and I write the spec/tests? :grinning:

lauraluiz:
:+1: but is it going to be in the JVM SDK?

schleichardt:
Yes, in the module sphere-convenience.

PR #1236, opened 2016-10-11
Allow retry just when there is a 409 response, issue #32

ccsalazarr-commercetools:


commercetools-convenience/src/main/java/io/sphere/sdk/sequencegenerators/CustomObjectBigIntegerNumberGenerator.java
```diff
@@ -53,14 +54,23 @@ private CustomObjectBigIntegerNumberGenerator(final CustomObjectBigIntegerNumber
...
             return CompletableFutureUtils.recoverWith(bigIntegerCompletionStage, (error) -> tryGetNextNumber(timeToLive - 1, error));
         } else {
             return CompletableFutureUtils.failed(throwable);
         }
     }
 
+    private boolean isNullOrConcurrentException( final Throwable throwable ) {
+        if ( throwable == null ){
+            return true;
+        }else if(throwable.getCause() instanceof ConcurrentModificationException) {
```
yanns: try to stay to the current code style:
```
    if (throwable == null) {
       ...
    } else if (...) {
    }
```

schleichardt:
Tests are missing.

ccsalazarr-commercetools:
@yanns and @schleichardt could you review the test (of the concurrent modification exception) to see if the approach is right, or if I have to do another kind of test.

yanns:
for me, it's ok. Let's wait for @schleichardt here.

commercetools-convenience/src/test/java/io/sphere/sdk/sequencegenerators/BigIntegerNumberGeneratorIntegrationTest.java
```diff
@@ -92,6 +100,44 @@ public void concurrentUsageTest() throws Exception {
...
+    public void failWithNotConcurrentModificationException() {
+        final String container = randomKey();
+        final String key = randomKey();
+        final CustomObjectBigIntegerNumberGeneratorConfig config =
+                CustomObjectBigIntegerNumberGeneratorConfigBuilder.of(getSphereClient(getBadRequestHttpClient()), key)
+                        .container(container)
+                        .build();
+        final BigIntegerNumberGenerator generator = CustomObjectBigIntegerNumberGenerator.of(config);
+        try {
+            generator.getNextNumber().toCompletableFuture().join();
```
schleichardt: If this does not throw an exception then the test would be still green.

Have a look at http://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html#exception-assertion and use `catchThrowable`.

schleichardt: The test does not validate that it does not retry, it still could retry a lot of times and finally failes with the exception.

commercetools-convenience/src/test/java/io/sphere/sdk/sequencegenerators/BigIntegerNumberGeneratorIntegrationTest.java
```diff
@@ -82,7 +90,7 @@ public void customObjectContainerAndKeyCanBeGiven() throws Exception {
     @Test
     public void concurrentUsageTest() throws Exception {
         final int firstNumber = 1;
-        final int lastNumber = 20;
+        final int lastNumber = 50;
```
schleichardt: Why did you change that?

commercetools-convenience/src/test/java/io/sphere/sdk/sequencegenerators/BigIntegerNumberGeneratorIntegrationTest.java
```diff
@@ -92,6 +101,44 @@ public void concurrentUsageTest() throws Exception {
...
+    public void failWithNotConcurrentModificationException() {
+        final String container = randomKey();
+        final String key = randomKey();
+        final CustomObjectBigIntegerNumberGeneratorConfig config =
+                CustomObjectBigIntegerNumberGeneratorConfigBuilder.of(getSphereClient(getBadRequestHttpClient()), key)
+                        .container(container)
+                        .build();
+        final BigIntegerNumberGenerator generator = CustomObjectBigIntegerNumberGenerator.of(config);
+
+        Throwable thrown = catchThrowable(() -> { generator.getNextNumber().toCompletableFuture().join(); });
```
schleichardt: this should also work: `catchThrowable(() -> generator.getNextNumber().toCompletableFuture().join());`

PR #1248, opened 2016-10-20
Remove inverted bounds exception, closes #1247

### LINKED ISSUE #1247: Range validations raise exception due to lexicographic comparison
Filter/Facet Range<T> compares the bounds to decide whether the range is valid. Otherwise it would raise a `InvertedBoundsException`. The problem arises when numbers are passed as strings, which are compared in a lexicographic manner instead of numerically, thus raising an exception for inverted bounds where it is actually correct from the numerical perspective.

Passing numbers as strings is unavoidable, since in most cases the desired bounds are obtained as text (e.g. query string). Therefore either the exception is removed (and let the developer take responsibility of the validity of his range) or an additional layer of complexity is added to always receive the correct type, forcing the developer to previously convert the string to whatever object.

lauraluiz:
to avoid lexicographic order on numerical values

@schleichardt please review.


commercetools-sdk-base/src/main/java/io/sphere/sdk/search/model/InvertedBoundsException.java
```diff
@@ -1,18 +0,0 @@
-package io.sphere.sdk.search.model;
-
-/**
- * Exception for the case that the lower bound is greater than the upper bound.
- *
- * {@include.example io.sphere.sdk.search.model.RangeTest#isInvalidWhenBoundsAreInverted()}
- *
- */
-public class InvertedBoundsException extends InvalidRangeException {
```
schleichardt: can we keep the exception? so at least previous code compiles

commercetools-sdk-base/src/main/java/io/sphere/sdk/search/model/Range.java
```diff
@@ -105,8 +102,8 @@ public boolean isBounded() {
 
     public String serialize(final Function<T, String> serializer) {
         return String.format("(%s to %s)",
-                Optional.ofNullable(lowerEndpoint()).map(serializer::apply).orElse(UNBOUND),
-                Optional.ofNullable(upperEndpoint()).map(serializer::apply).orElse(UNBOUND));
+                Optional.ofNullable(lowerEndpoint()).map(serializer).orElse(UNBOUND),
```
schleichardt: interesting

commercetools-sdk-base/src/test/java/io/sphere/sdk/search/model/RangeTest.java
```diff
@@ -64,13 +64,6 @@ public void isEmptyWhenItCannotContainAnyValue() throws Exception {
     }
 
     @Test
-    public void isInvalidWhenBoundsAreInverted() throws Exception {
```
schleichardt: I would also keep the test and specify that the exception is not thrown.

schleichardt:
I let travis check it and then merge.

schleichardt:
travis is failing, the example from the exceptions needs to be removed: https://travis-ci.org/commercetools/commercetools-jvm-sdk/builds/169230136#L1257

PR #1250, opened 2016-11-15
add code generation for shipping methods, zones and projects

schleichardt:
@lauraluiz please review, I suggest to look inside the generated classes ShippingMethodDraftDsl and ShippingMethodDraftBuilder


lauraluiz:
I don't like much that the resulting methods for Java keywords are `isDefault` and `_default`. Not sure about the builder's policy in the JVM SDK, but I would expect `isDefault()` for the getter and `default(boolean x)` for the setter. What happens then if you enable the getters by default?

schleichardt:
> I don't like much that the resulting methods for Java keywords are isDefault and _default. 

Provide an alternative, please. `default(boolean x)` is not possible because `default` is a keyword in Java.

It should be no problem to enable the getter for this builder.

PR #1252, opened 2016-11-16
allow flag to disable matching variants

schleichardt:
@lauraluiz please review, but it is not urgent

### DIFF
--- a/commercetools-models/src/main/java/io/sphere/sdk/producttypes/ProductTypeDraft.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/producttypes/ProductTypeDraft.java
@@ -14,8 +14,7 @@ import java.util.List;
  */
 @JsonDeserialize(as = ProductTypeDraftDsl.class)
 @ResourceDraftValue(factoryMethods = {
-        @FactoryMethod(parameterNames = {"key", "name", "description", "attributes"}),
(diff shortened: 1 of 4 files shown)


commercetools-models/src/main/java/io/sphere/sdk/producttypes/ProductTypeDraft.java
```diff
@@ -14,8 +14,7 @@
  */
 @JsonDeserialize(as = ProductTypeDraftDsl.class)
 @ResourceDraftValue(factoryMethods = {
-        @FactoryMethod(parameterNames = {"key", "name", "description", "attributes"}),
-        @FactoryMethod(methodName = "ofKey", parameterNames = {"key"})
```
lauraluiz: I guess is not important for this PR, but why did you remove the ofKey?

schleichardt: At the creation of this generated factory method I thought `key` is the only mandatory parameter. Then I used it and got exceptions due to missing data for product types. I allowed myself to remove this method again since it has not been released yet.

schleichardt:
@lauraluiz Can you check also the last commit?

lauraluiz:
Looks good! But difficult to tell if you missed any :D

PR #1259, opened 2016-11-21
activate aggregations

schleichardt:
@lauraluiz please review - this PR should be reviewed before the others

### DIFF
--- a/commercetools-models/src/test/java/io/sphere/sdk/products/search/ProductProjectionSearchModelIntegrationTest.java
+++ b/commercetools-models/src/test/java/io/sphere/sdk/products/search/ProductProjectionSearchModelIntegrationTest.java
@@ -32,7 +32,7 @@ public abstract class ProductProjectionSearchModelIntegrationTest extends Integr
     protected static ProductType productType;
 
     private static final boolean expectAggregations = Optional.ofNullable(System.getenv("EXPECT_AGGREGATIONS"))
-            .map("false"::equals).orElse(true);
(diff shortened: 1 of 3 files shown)


commercetools-models/src/test/java/io/sphere/sdk/products/search/ProductProjectionSearchModelIntegrationTest.java
```diff
@@ -32,7 +32,7 @@
     protected static ProductType productType;
 
     private static final boolean expectAggregations = Optional.ofNullable(System.getenv("EXPECT_AGGREGATIONS"))
-            .map("true"::equals).orElse(false);
+            .map("true"::equals).orElse(true);
```
lauraluiz: Here we say "if this var exists (and does not contain true) it will be false, otherwise by default is true". But it feels a bit weird because usually a var that contains something is `true` except for few cases (`false`, null, 0, empty...).

Of course the code becomes a bit more complex:
```
.map(env -> !env.equals("false")).orElse(true);
```

Well, it's a test anyway.

lauraluiz: AFAIK no

OlegIlyenko: I 'm not 100% sure, I think @peter-gerhard know it for sure

PR #1268, opened 2016-11-25
Add product type reference to line item

schleichardt:
there is still an issue in the platform to fix


lauraluiz:
Thanks @peter-gerhard !

PR #1274, opened 2016-11-29
Improve product attr docu

schleichardt:
@lauraluiz please review

* in this pull request I want to mention the DefaultProductAttributeFormatter in the product attribute documentation
* DefaultProductAttributeFormatter can also format the labels


commercetools-models/src/main/java/io/sphere/sdk/products/attributes/AttributeTableRow.java
```diff
@@ -0,0 +1,24 @@
...
+    @Nullable
+    String getTranslatedValue();
+
+    /**
+     * Returns the translated and formatted value of the label of the attribute.
+     *
+     * @return label
+     */
+    @Nullable
+    String getTranslatedLabel();
```
lauraluiz: `Translated` is misleading, as even in the comments say it's translated and formatted. Of course `getTranslatedAndFormattedLabel` is both long and ugly. One solution I can think of is to keep just `getLabel` and change the class name to reflect the translated/formatted thing.

commercetools-models/src/main/java/io/sphere/sdk/products/attributes/AttributeTable.java
```diff
@@ -0,0 +1,30 @@
...
+     * @return the maximal translated attribute value length or 0 in case of 0 rows
+     */
+    int getMaxTranslatedValueLength();
+
+    /**
+     * Iterates through the rows and finds the longest length of the translated label length.
+     *
+     * @return the maximal translated attribute label length or 0 in case of 0 rows
+     */
+    int getMaxTranslatedLabelLength();
```
lauraluiz: I think we are crossing a dangerous line here by introducing pure UI-focused elements into the SDK models. Into SDK convenience would be acceptable. Although I cannot really imagine a use case where we need to print the attributes in tables directly from the Java code, except maybe some console printing.

commercetools-internal-docs/src/main/java/io/sphere/sdk/meta/ProductAttributeDocumentation.java
```diff
@@ -104,7 +113,21 @@
...
+
+ {@include.example io.sphere.sdk.products.attributes.ProjectNameProductAttributeFormatter}
+
+ <p>The example in action:</p>
+
+ {@include.example io.sphere.sdk.products.attributes.DefaultProductAttributeFormatterDemo}
+
+
+
+ <h3 id="attribute-table-creation-pattern-matching">Creating a table of attributes (old way)</h3>
```
schleichardt: So should I remove it?

schleichardt: Hmm, it is very hard to use, I would always go with the new way. I don't see a use case.

schleichardt:
@lauraluiz updated for you

schleichardt:
Yes, it looks a bit weird, but that is the `Pair` or `Tuple2` class of Java SE. :roll_eyes:

PR #1277, opened 2016-12-08
replace SPHERE.IO logo with commercetools logo

schleichardt:


schleichardt:
@lauraluiz this is too huge, right? 
https://github.com/commercetools/commercetools-jvm-sdk/blob/4f866931a61ed1a87fda44198946875c2affe91d/README.md

lauraluiz:
Ehm... Yes. It even scared me a bit 😨

lauraluiz:
Actually I already tried that, I spent some long minutes trying to figure out how would it look best in Sunrise a logo from the press-kit. I think the only solution I found was to use directly HTML code, but I couldn't find a good look either (alineation and size) so that's why I didn't fix it yet. Wish you better luck :D

Issue #696, opened 2015-09-07
create subtotal helper method


schleichardt:
subtotal = sum(all line items)
Without cart discounts and shipping costs

schleichardt:
maybe we wait for the platform if they provide new fields

schleichardt:
implemented as io.sphere.sdk.carts.CartLike#calculateSubTotalPrice

PR #1282, opened 2016-12-14
Provide some methods to obtain information from the prices

lauraluiz:
@gogregor @OlegIlyenko @yanns 

I'm trying to provide some helpers that hide a bit whatever complexity might exist to perform habitual e-commerce operations. This is the same code appearing on most of the projects. Some methods may look quite small but they serve for making other calculations a lot simpler.

Gross/Net calculations on product cannot be performed in the JVM SDK because the TaxCategory belongs to the Product, which is unknown from the Variant level. And even if we would have it passed as a parameter, from the Product level it is not possible to decide which TaxRate needs to be applied without expanding the TaxCategory reference and selecting the correct TaxRate according to some country and state. A bunch of boilerplate code. Of course this boilerplate code needs then to be moved into every shop project, including Sunrise, which is not desirable from my point of view. The only solution I can find to be able to provide this functionality from the JVM SDK is to return the selected TaxRate along with the "selected" price (`price` or `scopePrice`). Any thoughts?

The fact that I cannot provide methods for products stops me a bit on providing the other helper methods, it feels unbalanced. But on the other side I prefer to have half the helpers than none.

I will target discounts and other calculations on future PR.


commercetools-models/src/test/java/io/sphere/sdk/carts/CartLikeTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+    private Cart cartWithTaxes;
+    private Cart cartWithoutTaxes;
+
+    @Before
+    public void setUp() throws Exception {
+        cartWithTaxes = SphereJsonUtils.readObjectFromResource("cart-with-taxes.json", Cart.class);
+        cartWithoutTaxes = SphereJsonUtils.readObjectFromResource("cart-without-taxes.json", Cart.class);
+    }
+
```
lauraluiz: These tests are not very easy to understand and verify, since the prices are extracted from the JSON file and that file is not precisely small. I'm not sure how to do it better, because the JVM SDK as it is now and without using any mocking framework, allows only either to provide your own implementation of the interface (as in PriceUtilsTest) or to parse a JSON file like in here. In the future I would like it to provide testing doubles.

Anyway, any idea on how to make it better?

lauraluiz: One option would be to leave in the JSON file only the minimum required info, but that has the risk of losing useful data for future tests or implementations of the methods (and generating again the data is time-costly).

commercetools-models/src/main/java/io/sphere/sdk/carts/CartLike.java
```diff
@@ -200,4 +203,88 @@ default CurrencyUnit getCurrency() {
...
+     */
+    default MonetaryAmount estimateSubTotalNetPrice() {
+        final MonetaryAmount lineItemTotal = getLineItems().stream()
+                .map(LineItem::estimateTotalNetPrice)
+                .reduce(zeroAmount(getCurrency()), MonetaryAmount::add);
+        final MonetaryAmount customLineItemTotal = getCustomLineItems().stream()
+                .map(CustomLineItem::estimateTotalNetPrice)
+                .reduce(zeroAmount(getCurrency()), MonetaryAmount::add);
+        return lineItemTotal.add(customLineItemTotal);
+    }
```
lauraluiz: https://github.com/commercetools/commercetools-jvm-sdk/issues/696

commercetools-models/src/test/java/io/sphere/sdk/products/PriceUtilsTest.java
```diff
@@ -0,0 +1,137 @@
...
+    @Test
+    public void returnsZeroAmount() throws Exception {
+        final MonetaryAmount zeroAmount = zeroAmount(EUR);
+        assertThat(zeroAmount.getNumber().doubleValue()).isEqualTo(0);
+        assertThat(zeroAmount.getCurrency()).isEqualTo(EUR);
+    }
+
+    @Test
+    public void convertsToGross() throws Exception {
+        final MonetaryAmount netAmount = monetaryAmountOf(100);
```
yanns: Maybe you can use more complicated numbers to check the rounding issues?

commercetools-models/src/test/java/io/sphere/sdk/carts/CartLikeTest.java
yanns: Is there any reason not to use a mocking framework?

Can you use the `*Impl` for the tests?
Maybe you could load a "base" cart with json and set the fields relevant for the tests in java.
(you may have to use `instanceof` and `*Impl` classes for that - not particularly pretty)

Otherwise I'd go with minimizing the json files.

commercetools-models/src/main/java/io/sphere/sdk/carts/CartLike.java
```diff
@@ -200,4 +203,88 @@ default CurrencyUnit getCurrency() {
...
+     * be determined whether taxes are included or not in the calculation, as it depends on the tax categories
+     * linked to the items in cart and the country/state of the cart.
+     *
+     * @return the estimated total gross price of the cart
+     */
+    default MonetaryAmount estimateTotalGrossPrice() {
+        return Optional.ofNullable(getTaxedPrice())
+                .map(TaxedPrice::getTotalGross)
+                .orElseGet(this::getTotalPrice);
+    }
```
OlegIlyenko: I feel that throwing an exception or `Oprtional.absent` is a better choice if we don't know yet whether taxes are included or not. I find it quite confusing if `estimateTotalGrossPrice` will return a net price in some circumstances.

commercetools-models/src/test/java/io/sphere/sdk/carts/CartLikeTest.java
lauraluiz: Regarding the mocking framework, it's just a decision that was took long time ago that in general you should be able to test the JVM SDK without mocking (which was not possible with the previous Play SDK). Also in many projects we found tests that ended up not testing anything because of a misuse of the mocking capability. All together we decided to eat our own dog food and not use mocking frameworks. Which in the SDK is not a big deal because mostly everything are IT. But in Sunrise and other places, it can become really annoying to test and I consider the mocking framework solution.

I "can" use Impl but it's really not an option because it's just a huge constructor with all possible values coming from the API, not useful at all.

The only solution, besides the mocking framework, is to auto-generate test doubles, but that's actually just an equivalent of mocking.

commercetools-models/src/test/java/io/sphere/sdk/products/PriceUtilsTest.java
lauraluiz: I considered that but the results could be a bit confusing and I forgot to add a comment here.

The problem is that the Moneta implementation of Money is not rounding (which is good actually because we want to keep the most precise value until the very end) but then some  prices might be returned as `10.33333333333...` which looks weird in the tests.

WDYT?

lauraluiz:
So we decided to drop the gross/net calculations. There are many scenarios, many more than we probably think, about tax calculations. Too complex to provide all of them from the SDK level.

Better to provide some "dummy" methods that help the developer to get enough info to calculate the gross/net with the PriceUtils class.

Some of the methods we can consider implementing are:
```
- Sum of all discounted prices applied to the line item
- Sum of all discounted prices applied to the line item per cart discount
- Cart subtotal
- Checker of cart integrity (e.g. no mix of gross and net)
- Utils to calculate gross/net without any logic
- Help to find correct tax rate for product level
```
I will then remove the gross/net variants and keep or adapt the rest :)

commercetools-models/src/test/java/io/sphere/sdk/carts/CartLikeTest.java
lauraluiz: I have minimized the JSON files to only pricing related info. The one without taxes looks quite okay, the other one is... well. Actually in this particular case I find it useful that I'm forced to provide the JSON from real CTP projects directly. It's easy to mistake some calculation or assume a certain price configuration.

commercetools-models/src/test/java/io/sphere/sdk/products/PriceUtilsTest.java
lauraluiz: I've added two tests to check it's actually not rounding.

lauraluiz:
Okay, ready again to review 🙌 
PriceUtils didn't change, only its tests.

commercetools-models/src/test/java/io/sphere/sdk/products/PriceUtilsTest.java
yanns: 👍

commercetools-models/src/test/resources/carts/cart-with-taxes.json
```diff
@@ -0,0 +1,158 @@
...
+        },
+        "totalGross": {
+          "currencyCode": "EUR",
+          "centAmount": 8400
+        }
+      }
+    }
+  ],
+  "taxMode": "Platform"
+}
```
yanns: good example!

lauraluiz:
@gogregor any objection to merging? :)

commercetools-models/src/main/java/io/sphere/sdk/products/PriceUtils.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /**
+     * Converts the given net amount (i.e. without taxes) to gross (i.e. with taxes included) according to the provided tax rate.
+     * @param netAmount the net monetary amount
+     * @param taxRate the given tax rate, e.g. {@code 0.19} for 19% tax
+     * @return the gross monetary amount
+     */
+    public static MonetaryAmount convertNetToGrossPrice(final MonetaryAmount netAmount, final double taxRate) {
+        return netAmount.multiply(1 + taxRate);
```
gogregor: 👍

commercetools-models/src/main/java/io/sphere/sdk/products/PriceUtils.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /**
+     * Converts the given gross amount (i.e. with taxes included) to net (i.e. without taxes) according to the provided tax rate.
+     * @param grossAmount the gross monetary amount
+     * @param taxRate the given tax rate, e.g. {@code 0.19} for 19% tax
+     * @return the net monetary amount
+     */
+    public static MonetaryAmount convertGrossToNetPrice(final MonetaryAmount grossAmount, final double taxRate) {
+        return grossAmount.divide(1 + taxRate);
```
gogregor: 👍

PR #1287, opened 2016-12-27
Fix Revapi report on covariant return type changes

lauraluiz:
And document latest changes accordingly


lauraluiz:
Temporary fix until `revapi-java-0.12.0` released (see https://github.com/revapi/revapi/issues/56#issuecomment-269135425)

PR #1292, opened 2017-01-09
Add support for category assets

nikola-mladenovic:
@lauraluiz - LMK how does this look to you, and if it makes sense we can create an issue in the commercetools/commercetools-jvm-sdk project.

### DIFF
--- a/commercetools-models/src/main/java/io/sphere/sdk/categories/Category.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/categories/Category.java
@@ -175,6 +175,7 @@ public interface Category extends Resource<Category>, WithLocalizedSlug, MetaAtt
     CustomFields getCustom();
 
     @Nonnull
+    @IgnoreInQueryModel
     List<Asset> getAssets();
 
     @Override
--- a/commercetools-models/src/main/java/io/sphere/sdk/categories/Category.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/categories/Category.java
@@ -11,6 +11,7 @@ import io.sphere.sdk.types.CustomFields;
 import io.sphere.sdk.types.TypeDraft;
 import org.apache.commons.lang3.builder.ReflectionToStringBuilder;
 
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.util.List;
 
@@ -172,6 +173,9 @@ public interface Category extends Resource<Category>, WithLocalizedSlug, MetaAtt
      */
     @Nullable
     CustomFields getCustom();
+
+    @Nonnull
+    List<Asset> getAssets();
 
     @Override
     default Reference<Category> toReference() {
--- a/commercetools-models/src/main/java/io/sphere/sdk/categories/commands/updateactions/AddAsset.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/categories/commands/updateactions/AddAsset.java
@@ -1,0 +1,21 @@
+package io.sphere.sdk.categories.commands.updateactions;
+
+import io.sphere.sdk.categories.Category;
+import io.sphere.sdk.commands.UpdateActionImpl;
+import io.sphere.sdk.models.AssetDraft;
+
+/**
+ * Adds an asset.
+ */
+public final class AddAsset extends UpdateActionImpl<Category> {
+    private final AssetDraft asset;
+
+    private AddAsset(final AssetDraft asset) {
+        super("addAsset");
+        this.asset = asset;
+    }
+
+    public AssetDraft getAsset() {
+        return asset;
+    }
+}
--- a/commercetools-models/src/main/java/io/sphere/sdk/categories/commands/updateactions/AddAsset.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/categories/commands/updateactions/AddAsset.java
@@ -18,4 +18,8 @@ public final class AddAsset extends UpdateActionImpl<Category> {
     public AssetDraft getAsset() {
         return asset;
     }
+
+    public static AddAsset of(final AssetDraft asset) {
+        return new AddAsset(asset);
+    }
 }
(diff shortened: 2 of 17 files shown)


commercetools-models/src/main/java/io/sphere/sdk/categories/commands/updateactions/SetAssetCustomField.java
```diff
@@ -0,0 +1,32 @@
...
+    private SetAssetCustomField(final String assetId, final String name, final JsonNode value) {
+        super("setAssetCustomField", name, value);
+        this.assetId = assetId;
+    }
+
+    public String getAssetId() {
+        return assetId;
+    }
+
+    public static SetAssetCustomField usingJson(final String assetId, final String name, final JsonNode value) {
```
lauraluiz: Now I'm checking the same with products, I guess you removed it because it sounded weird. Let's see... I would just simply say `ofJsonValue`, sounds better than `ofUsingJson` or `ofJson` (which is misleading). The "using" part from the other classes forces us to change a bit the structure here, which is bad, but it's the best option I can think of and yet it's easy to understand.

commercetools-models/src/main/java/io/sphere/sdk/categories/Category.java
```diff
@@ -173,6 +174,10 @@
     @Nullable
     CustomFields getCustom();
 
+    @Nonnull
```
lauraluiz: Okay just saw that Michael used it here for Products... I wonder why. But Products is not auto-generated and I don't see anything else problematic here, so give it a try.

lauraluiz: Anyway, leave it. We found the issue elsewhere :)

commercetools-models/src/main/java/io/sphere/sdk/categories/commands/updateactions/SetAssetCustomField.java
nikola-mladenovic: I also thought about `ofJsonValue` - it lead me to think the `final JsonNode value` is the first param, and not the `final String assetId`, but perhaps that's just me, coming from the Swifty background. Let's use `ofJsonValue`, cause I don't think we'll figure out anything better here ☺️

lauraluiz: Yep. I share your view: not perfect but probably the best ;)

commercetools-models/src/main/java/io/sphere/sdk/categories/commands/updateactions/AddAsset.java
```diff
@@ -0,0 +1,21 @@
...
+    private final AssetDraft asset;
+
+    private AddAsset(final AssetDraft asset) {
+        super("addAsset");
+        this.asset = asset;
+    }
+
+    public AssetDraft getAsset() {
+        return asset;
+    }
```
nikola-mladenovic: 👍 Added a static constructor method for `AddAsset` update action.

nikola-mladenovic:
I've updated `CategoryBuilder.java` and added setter for `assets`: `public CategoryBuilder assets(final List<Asset> assets)`, so it can be used if needed for `Category` tests.

Now the project builds properly, but I cannot run all those integration tests because of missing project credentials.

@lauraluiz - if & what else do you think we need here before merging?

nikola-mladenovic:
Added tests which cover category assets, there's one small outstanding issue on the API side to be addressed for `assetCustomType` test. After that's been fixed, we could go ahead with the merge.

commercetools-models/src/main/java/io/sphere/sdk/categories/commands/updateactions/ChangeAssetOrder.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.util.List;
+
+/**
+ * Changes the order of the assets list. The new order is defined by listing the ids of the assets.
+ */
+public final class ChangeAssetOrder extends UpdateActionImpl<Category> {
+    private final List<String> assetOrder;
+
+    public ChangeAssetOrder(final List<String> assetOrder) {
```
lauraluiz: Constructor should be private and there should be a static constructor available (`of` method).

commercetools-models/src/main/java/io/sphere/sdk/categories/commands/updateactions/RemoveAsset.java
```diff
@@ -0,0 +1,20 @@
...
+import io.sphere.sdk.categories.Category;
+import io.sphere.sdk.commands.UpdateActionImpl;
+
+/**
+ * Removes an asset.
+ */
+public final class RemoveAsset extends UpdateActionImpl<Category> {
+    private final String assetId;
+
+    public RemoveAsset(final String assetId) {
```
lauraluiz: Same as `ChangeAssetOrder`: constructor private and creation via static constructor.

commercetools-models/src/main/java/io/sphere/sdk/categories/commands/updateactions/SetAssetDescription.java
```diff
@@ -0,0 +1,35 @@
+package io.sphere.sdk.categories.commands.updateactions;
+
+import io.sphere.sdk.categories.Category;
+import io.sphere.sdk.commands.UpdateActionImpl;
+import io.sphere.sdk.models.LocalizedString;
+
+import javax.annotation.Nullable;
+
+/**
+ * Changes the name of an asset.
```
lauraluiz: *"Changes the description"
By the way, could you fix this same mistake from the Product's action `SetAssetDescription`? ;)

nikola-mladenovic:
Addressed all remarks ☺️

nikola-mladenovic:
@lauraluiz - CI is finally green, feel free to merge when you get a chance ☺️

PR #1299, opened 2017-01-18
Make Attribute.getValueAsStringSet() preserve input order,  close #1294

### LINKED ISSUE #1294: Attribute.getValueAsStringSet() does not preserve order of values
Hi,

The documentation for product attributes of type "Set" states: "The set attribute type defines a set (array without duplicates) with values of the given elementType" (http://dev.commercetools.com/http-api-projects-productTypes.html#settype). Since this definition is unclear about the order of values in the Set, we asked commercetools support a while ago. They said the order is retained as in any proper array.

However, this seems to be wrong as far as it concerns the jvm-sdk. The following Test fails even with sdk version 1.9.0:

` final String exampleProductJson = " {\n" +
               {...}
                "      \"masterData\": {\n" +
                "        \"current\": {\n" +
               {...}
                "          \"masterVariant\": {\n" +
               {...}
                "            \"attributes\": [\n" +
                "              {\n" +
                "                \"name\": \"testAttribute\",\n" +
                "                \"value\": [" +
                "                    \"3\",\n" +
                "                  \"1\",\n" +
                "                  \"2\"" +
                "              }\n" +
                "            ],\n" +
               {...}
                "    }";`

           ` final Product parsedProduct = SphereJsonUtils.readObject(exampleProductJson, Product.class);
            assertThat(parsedProduct.getMasterData().getCurrent().getMasterVariant().getAttribute("testAttribute").getValueAsStringSet())
                .containsExactly("3", "1","2");`

In contrast to the Admin-Center: here the attribute values are shown in the correct order.
Would be nice if we could see this fixed in the next sdk version.

Thank you!

ccsalazarr-commercetools:
@lauraluiz could you review it please?

### DIFF
--- a/commercetools-sdk-base/src/main/java/io/sphere/sdk/json/SphereJsonUtils.java
+++ b/commercetools-sdk-base/src/main/java/io/sphere/sdk/json/SphereJsonUtils.java
@@ -3,6 +3,7 @@ package io.sphere.sdk.json;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.*;
+import com.fasterxml.jackson.databind.module.SimpleModule;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.databind.type.TypeFactory;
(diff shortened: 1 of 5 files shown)


commercetools-sdk-base/src/main/java/io/sphere/sdk/json/SphereJsonUtils.java
```diff
@@ -131,12 +136,11 @@ public static String prettyPrint(final JsonNode jsonNode) {
...
 
     /**
-     *
      * Reads a UTF-8 JSON text file from the classpath of the current thread and transforms it into a Java object.
      *
-     * @param resourcePath the path to the resource. Example: If the file is located in "src/test/resources/foo/bar/product.json" then the path should be "foo/bar/product.json"
+     * @param resourcePath  the path to the resource. Example: If the file is located in "src/test/resources/foo/bar/product.json" then the path should be "foo/bar/product.json"
      * @param typeReference the full generic type information about the object to create
-     * @param <T> the type of the result
+     * @param <T>           the type of the result
```
lauraluiz: I'm starting to think your IDE played with the Javadoc :D

ccsalazarr-commercetools: @lauraluiz you are right, it was IntelliJ who changes the documentation, it happened when I used the command to format the code. I will keep it as before :)

PR #1307, opened 2017-02-08
Javadoc 8u121 fix, closes #1303

### LINKED ISSUE #1303: Javadoc generation fails with JDK >= 8u121
I'm currently using the latest JDK 1.8.0_121 and the maven build fails with the following error:

[ERROR] javadoc: error - Argument for -bottom contains JavaScript.
[ERROR] Use --allow-script-in-comments to allow use of JavaScript.

The --allow-script-in-comments javadoc option was added with JDK 1.8.0_121 (see: http://www.oracle.com/technetwork/java/javase/8u121-relnotes-3315208.html).

We should add it to our maven build when a JDK >= is used for building.

katmatt:


tolbertam:
Just curious because I'm bumping into the same issue in another project and you probably came to the same conclusion as me,  JDK profile activation with ranges does not work, i.e.:

```xml
<activation>
    <jdk>[1.8.0_121,)</jdk>
</activation>
```

fails because the [JDK profile activator code](https://github.com/apache/maven/blob/master/maven-model-builder/src/main/java/org/apache/maven/model/profile/activation/JdkVersionProfileActivator.java#L133-L141) does not account for update versions for ranges, so I end up getting a `NumberFormatException`:

```
[ERROR]   
[ERROR]   The project xxxxxx:0.0.1-SNAPSHOT (xxxxx/pom.xml) has 1 error
[ERROR]     Failed to determine activation for profile doclint-java8-disable @ line 89, column 18: For input string: "0_121"
java.lang.NumberFormatException: For input string: "0_121"
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Integer.parseInt(Integer.java:580)
	at java.lang.Integer.parseInt(Integer.java:615)
	at org.apache.maven.model.profile.activation.JdkVersionProfileActivator.getRelationOrder(JdkVersionProfileActivator.java:141)
	at org.apache.maven.model.profile.activation.JdkVersionProfileActivator.isInRange(JdkVersionProfileActivator.java:108)
	at org.apache.maven.model.profile.activation.JdkVersionProfileActivator.isActive(JdkVersionProfileActivator.java:79)
```

Without being able to specify a range, won't builds start failing for versions after 1.8.0_121?  Any plans for working around that?

katmatt:
Hi @tolbertam,

You are right that our workaround will break with future updates of JDK 1.8.0. But as long as our team and our build server don't use a JDK version > 18.0_121 we will keep our current solution. 

And as mentioned in [https://github.com/commercetools/commercetools-jvm-sdk/blob/master/pom.xml#L428], you can still activate the workaround manually. You could even add it to your `MAVEN_OPTS` environment variable, but I haven't tried that yet.

tolbertam:
Thanks @katmatt , I agree that this is probably the best approach to take, the capability to activate the profile via command line sounds good.

PR #1310, opened 2017-02-10
Shopping lists

katmatt:


commercetools-models/src/main/java/io/sphere/sdk/shoppinglists/LineItem.java
```diff
@@ -0,0 +1,54 @@
...
+import io.sphere.sdk.producttypes.ProductType;
+import io.sphere.sdk.types.CustomFields;
+import io.sphere.sdk.types.TypeDraft;
+
+import javax.annotation.Nullable;
+import java.time.ZonedDateTime;
+
+@JsonDeserialize(as = LineItemImpl.class)
+@ResourceValue
+public interface LineItem {
```
lauraluiz: Can you write here some documentation like the one that cart's line item has? 

Something like:
```
A line item is a reference to a product variant in a product in its current version.

A LineItem can have {@link io.sphere.sdk.types.Custom custom fields}.

 @see io.sphere.sdk.shoppinglists.commands.updateactions.AddLineItem
 @see io.sphere.sdk.shoppinglists.commands.updateactions.RemoveLineItem
 @see io.sphere.sdk.shoppinglists.commands.updateactions.ChangeLineItemQuantity
 @see ShoppingList#getLineItems()
```

I know it's difficult to keep up to date, so we can discuss it later too.

lauraluiz: Ah and probably the other classes need it too.

commercetools-models/src/main/java/io/sphere/sdk/shoppinglists/ShoppingListDraft.java
```diff
@@ -0,0 +1,40 @@
...
+    @Nullable
+    String getKey();
+
+    @Nullable
+    Reference<Customer> getCustomer();
+
+    @Nullable
+    LocalizedString getSlug();
+
+    List<LineItemDraft> getLineItems();
```
lauraluiz: In the docu says it's optional, but I had a case too in the past where in an array it said optional when it was never the case. Have you checked that it's really never null, even when the shopping list is newly created? If in that case it's returning an empty array, then we have to report it to the docu so that they fix it.

commercetools-models/src/main/java/io/sphere/sdk/shoppinglists/commands/updateactions/AddLineItem.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public String getProductId() {
+        return productId;
+    }
+
+    public Integer getVariantId() {
+        return variantId;
+    }
+
+    public Long getQuantity() {
```
lauraluiz: Missing `@Nullable` annotation

commercetools-models/src/main/java/io/sphere/sdk/shoppinglists/commands/updateactions/AddLineItem.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public static AddLineItem of(final String productId, final int variantId, final Long quantity) {
+        return new AddLineItem(productId, variantId, quantity, null, null);
+    }
+
+    public String getProductId() {
+        return productId;
+    }
+
+    public Integer getVariantId() {
```
lauraluiz: Missing `@Nullable` annotation

commercetools-models/src/main/java/io/sphere/sdk/shoppinglists/commands/updateactions/AddLineItem.java
```diff
@@ -0,0 +1,80 @@
...
+        this.quantity = quantity;
+        this.addedAt = addedAt;
+        this.custom = custom;
+    }
+
+    public static AddLineItem of(final ProductIdentifiable product, final int variantId, final Long quantity) {
+        return of(product.getId(), variantId, quantity);
+    }
+
+    public static AddLineItem of(final String productId, final int variantId, final Long quantity) {
```
lauraluiz: Quantity is optional here, so we should annotate it and provide a constructor without quantity too, to avoid users to have to explicitly pass `null`.

lauraluiz: Oh and variant too. I would keep a constructor with product, variant and quantity together, for keeping the similarity with cart's line item, but I would provide one constructor with only product and enable `with` methods for the rest.

commercetools-models/src/main/java/io/sphere/sdk/shoppinglists/commands/updateactions/AddTextLineItem.java
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * {@doc.gen intro}
+ *
+ * {@include.example io.sphere.sdk.shoppinglists.commands.ShoppingListUpdateCommandIntegrationTest#addTextLineItem()}
+ *
+ * @see ShoppingList#getTextLineItems()
+ */
+public final class AddTextLineItem extends UpdateActionImpl<ShoppingList> implements CustomDraft {
+    private final LocalizedString name;
+    private final Long quantity;
```
lauraluiz: Quantity is optional here too, so the getters and constructors should all have the nullable annotation, plus the constructor need to be adapted accordingly.

commercetools-models/src/main/java/io/sphere/sdk/shoppinglists/commands/updateactions/AddTextLineItem.java
```diff
@@ -0,0 +1,66 @@
...
+        this.description = description;
+        this.custom = custom;
+    }
+
+    public static AddTextLineItem of(final LocalizedString name, final long quantity) {
+        return new AddTextLineItem(name, quantity, null, null);
+    }
+
+    @Nullable
+    public LocalizedString getName() {
```
lauraluiz: Name is actually required ;)

commercetools-models/src/main/java/io/sphere/sdk/shoppinglists/commands/updateactions/ChangeName.java
```diff
@@ -0,0 +1,31 @@
...
+public final class ChangeName extends UpdateActionImpl<ShoppingList> {
+    private final LocalizedString name;
+
+    private ChangeName(final LocalizedString name) {
+        super("changeName");
+        this.name = name;
+    }
+
+    public static io.sphere.sdk.shoppinglists.commands.updateactions.ChangeName of(final LocalizedString name) {
+        return new io.sphere.sdk.shoppinglists.commands.updateactions.ChangeName(name);
```
lauraluiz: This can be imported here, I don't see any conflicts.

commercetools-models/src/main/java/io/sphere/sdk/shoppinglists/commands/updateactions/SetCustomer.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public static SetCustomer of(@Nullable final Referenceable<Customer> customer) {
+        final ResourceIdentifier<Customer> resourceIdentifier = Optional.ofNullable(customer)
+                .map(ResourceIdentifiable::toResourceIdentifier)
+                .orElse(null);
+        return new SetCustomer(resourceIdentifier);
+    }
+
+    @Nullable
+    public ResourceIdentifier<Customer> getCustomer() {
```
lauraluiz: This made me realize that sometimes in the docu something was mentioned as a `ReferenceIdentifier` when in here it was being used a `Reference`. I'm not sure what the difference is, but please find out and make sure that elsewhere we are returning the correct reference.

commercetools-models/src/main/java/io/sphere/sdk/shoppinglists/commands/updateactions/SetDescription.java
```diff
@@ -0,0 +1,31 @@
...
+ * Sets the description of the shopping list.
+ *
+ * {@doc.gen intro}
+ *
+ * {@include.example io.sphere.sdk.shoppinglists.commands.ShoppingListUpdateCommandIntegrationTest#setDescription()}
+ *
+ * @see ShoppingList#getDescription()
+ */
+public final class SetDescription extends UpdateActionImpl<ShoppingList> {
+    private final LocalizedString description;
```
lauraluiz: This is nullable (getter, constructor too)

commercetools-models/src/main/java/io/sphere/sdk/shoppinglists/commands/updateactions/SetKey.java
```diff
@@ -0,0 +1,30 @@
...
+ * Sets the key of the shopping list.
+ *
+ * {@doc.gen intro}
+ *
+ * {@include.example io.sphere.sdk.shoppinglists.commands.ShoppingListUpdateCommandIntegrationTest#setKey()}
+ *
+ * @see ShoppingList#getKey()
+ */
+public final class SetKey extends UpdateActionImpl<ShoppingList> {
+    private final String key;
```
lauraluiz: Nullable too (getter, constructor)

commercetools-models/src/main/java/io/sphere/sdk/shoppinglists/commands/updateactions/SetSlug.java
```diff
@@ -0,0 +1,31 @@
...
+ * Sets the slug of the shopping list.
+ *
+ * {@doc.gen intro}
+ *
+ * {@include.example io.sphere.sdk.shoppinglists.commands.ShoppingListUpdateCommandIntegrationTest#setSlug()}
+ *
+ * @see ShoppingList#getSlug()
+ */
+public final class SetSlug extends UpdateActionImpl<ShoppingList> {
+    private final LocalizedString slug;
```
lauraluiz: Nullable (getter, constructor)

commercetools-models/src/main/java/io/sphere/sdk/shoppinglists/commands/updateactions/SetTextLineItemDescription.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * {@doc.gen intro}
+ *
+ * {@include.example io.sphere.sdk.shoppinglists.commands.ShoppingListUpdateCommandIntegrationTest#setTextLineItemDescription()}
+ *
+ * @see ShoppingList#getTextLineItems()
+ */
+public final class SetTextLineItemDescription extends UpdateActionImpl<ShoppingList> {
+    private final String textLineItemId;
+    private final LocalizedString description;
```
lauraluiz: Nullable (getter, constructor)

commercetools-models/src/test/java/io/sphere/sdk/shoppinglists/queries/ShoppingListExpansionModelIntegrationTest.java
```diff
@@ -0,0 +1,75 @@
...
+public class ShoppingListExpansionModelIntegrationTest extends IntegrationTest {
+
+    @Test
+    public void lineItemsProductSlug() {
+        withTaxedProduct(client(), product -> {
+            final ShoppingListDraftDsl shoppingListDraft = newShoppingListDraftWithLineItems(product);
+            withShoppingList(client(), shoppingListDraft, shoppingList -> {
+                final ExpansionPathContainer<ShoppingList> productSlugExpansion = ShoppingListExpansionModel.of().lineItems().productSlug();
+                final ShoppingList fetchedShoppingList = client().executeBlocking(ShoppingListByIdGet.of(shoppingList.getId()).
+                        withExpansionPaths(productSlugExpansion));
```
yanns: good to see that you could implement the expansion on a non-referenced field!

commercetools-models/src/main/java/io/sphere/sdk/shoppinglists/LineItem.java
```diff
@@ -0,0 +1,54 @@
...
+    String getId();
+
+    String getProductId();
+
+    @Nullable
+    Integer getVariantId();
+
+    @Nullable
+    Reference<ProductType> getProductType();
```
yanns: no it's not optional.
It's optional on carts because we had to stay backwards compatible with the existing persisted data.
For a green field endpoint, I choose to have this field non optional because there is no reason for this.

commercetools-models/src/main/java/io/sphere/sdk/shoppinglists/ShoppingListDraft.java
yanns: I think it's possible to omit the field completely, and it should be possible to send `lineItems: null`. Please tell me if it is not the case.
We change the JSON parser a few months ago when Michael found out that the behaviour was different if the field was omitted or if the field was set the `null`. Now it should behave the same in both cases.

commercetools-models/src/main/java/io/sphere/sdk/shoppinglists/commands/updateactions/SetCustomer.java
yanns: A `ReferenceIdentifier` allows to use the `id` or the `key` of the customer.
A `Reference` only allows to use the `id` of the customer.

commercetools-models/src/main/java/io/sphere/sdk/carts/commands/updateactions/AddShoppingList.java
```diff
@@ -0,0 +1,56 @@
...
+import io.sphere.sdk.carts.Cart;
+import io.sphere.sdk.channels.Channel;
+import io.sphere.sdk.commands.UpdateActionImpl;
+import io.sphere.sdk.models.Reference;
+import io.sphere.sdk.shoppinglists.ShoppingList;
+
+import javax.annotation.Nullable;
+
+/**
+ * Adds a shopping list to a cart.
```
yanns: Do we link to the online docu?

commercetools-models/src/main/java/io/sphere/sdk/shoppinglists/commands/updateactions/ChangeTextLineItemQuantity.java
```diff
@@ -23,11 +23,11 @@ private ChangeTextLineItemQuantity(final String textLineItemId, final Long quant
         this.quantity = quantity;
     }
 
-    public static ChangeTextLineItemQuantity of(final String textLineItemId, final Long quantity) {
+    public static ChangeTextLineItemQuantity of(final String textLineItemId, final long quantity) {
```
lauraluiz: All primitive types were removed before v1.0.0. The reason was that sometimes a field became nullable (documentation error, evolution of the API...) and then the signature needed to change to the Object version, thus generating breaking changes for such a trivial thing.
Having a mix of primitive and objects depending on the resource felt a bit inconsistent too, but that's minor.

We could accept primitive types, keep a primitive type as field and return an Object. This way we gain in performance and yet keeping it breaking-changes-free. But we have to be careful to always keep it like this. But right now there is no gain in this code because we still keep the class field as Object.

So what do you think?

lauraluiz:
Looks good now! Let's only check tomorrow the primitive thingy :)

yanns:
🏆 🎉

selo-ecube:
The shopping lists sound interesting! Do you plan to release it soon?

lauraluiz:
@selo-ecube we are starting the release process now, if everything goes well it might be live late today :)

PR #1317, opened 2017-02-20
Channel geolocation

### LINKED ISSUE #1315: Add Channel.geoLocation support
We should add support for the `geoLocation` field of the channel (http://dev.commercetools.com/http-api-projects-channels.html#channel) tour SDK.

katmatt:
closes #1315


commercetools-sdk-base/src/main/java/io/sphere/sdk/models/GeoJSON.java
```diff
@@ -0,0 +1,15 @@
+package io.sphere.sdk.models;
+
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+/**
+ * Base class for GeoJSON objects that provides the required annotation
+ * to map to the correct sub type via the <code>type</code> property.
```
lauraluiz: Let me know if it's wrong, but shouldn't it be `{@code type}`, which behaves similarly in the generated Javadoc but belongs to the Javadoc specification? Sometimes we can't help it and we have to use HTML, but I would imagine this one is not the case.

PR #1322, opened 2017-02-24
Refactor exists and missing filter

katmatt:
This PR fixes the type hierarchy for the `exists` and `missing` predicates we discussed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,3 @@ notifications:
   email:
     on_success: change
     on_failure: change
-branches:
-  only:
-    - master
-    - generate-query-models
--- a/commercetools-sdk-base/src/main/java/io/sphere/sdk/search/model/TermFilterExistsAndMissingSearchModel.java
+++ b/commercetools-sdk-base/src/main/java/io/sphere/sdk/search/model/TermFilterExistsAndMissingSearchModel.java
@@ -6,7 +6,7 @@ import java.util.List;
 import java.util.function.Function;
 
 /**
- * Model to build term filters with exits and missing predicates
+ * Model to build term filters with exits and missing predicates.
  * .
  * @param <T> type of the resource
  * @param <V> type of the value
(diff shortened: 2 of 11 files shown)


.travis.yml
```diff
@@ -14,7 +14,3 @@ notifications:
   email:
     on_success: change
     on_failure: change
-branches:
-  only:
-    - master
-    - generate-query-models
```
lauraluiz: JFYI: This was done to speed up CI, which I guess that's not a problem for you ;)

commercetools-sdk-base/src/main/java/io/sphere/sdk/search/model/TermFilterExistsAndMissingSearchModel.java
```diff
@@ -0,0 +1,44 @@
...
+    List<FilterExpression<T>> containsAny(final Iterable<V> values);
+
+    @Override
+    List<FilterExpression<T>> containsAnyAsString(final Iterable<String> values);
+
+    @Override
+    List<FilterExpression<T>> is(final V value);
+
+    @Override
+    List<FilterExpression<T>> isIn(final Iterable<V> values);
```
lauraluiz: Is it to keep the highlighting in IDEs?

lauraluiz: If that's the case, why not the exists/missing ones?

katmatt: I wasn't sure why we repeat overridden methods when we extend some interfaces and tried to follow our code style. The only reason I had on my mind was that we did it so that javadoc copies the doc from the extended interface. But it does it automatically for non-public interfaces and for public interfaces you can jump to the javadoc of the super interfaces. 

Then I would suggest to remove these overriden methods when working on such an interface again 👍

lauraluiz: As said in person, there are two reasons why you would like to override methods in an interface or abstract class:
1) to show the most important methods in bold in the IDE
2) to rename an argument, in case you can provide a better name in this context

PR #1325, opened 2017-03-01
Generate abstract builder

katmatt:


katmatt:
@lauraluiz Please review my updated changes.

PR #1327, opened 2017-03-02
Add missing custom field in draft, impl and builder of orderImport. Close #792

### LINKED ISSUE #792: Add missing fields to OrderImport
like custom fields

http://dev.sphere.io/http-api-projects-orders-import.html

ccsalazarr-commercetools:
Add missing custom field in draft, impl and builder of orderImport. Close #792

### DIFF
--- a/commercetools-models/src/test/java/io/sphere/sdk/orders/OrderImportDraftBuilderTest.java
+++ b/commercetools-models/src/test/java/io/sphere/sdk/orders/OrderImportDraftBuilderTest.java
@@ -1,0 +1,99 @@
+package io.sphere.sdk.orders;
+
+import com.neovisionaries.i18n.CountryCode;
+import io.sphere.sdk.carts.InventoryMode;
+import io.sphere.sdk.carts.RoundingMode;
+import io.sphere.sdk.carts.TaxPortion;
+import io.sphere.sdk.carts.TaxedPrice;
+import io.sphere.sdk.customergroups.CustomerGroup;
+import io.sphere.sdk.models.Address;
+import io.sphere.sdk.models.LocalizedEnumValue;
+import io.sphere.sdk.models.LocalizedString;
+import io.sphere.sdk.models.Reference;
+import io.sphere.sdk.products.Price;
+import io.sphere.sdk.products.attributes.AttributeImportDraft;
+import io.sphere.sdk.taxcategories.TaxCategory;
+import io.sphere.sdk.types.CustomFieldsDraft;
+import io.sphere.sdk.types.CustomFieldsDraftBuilder;
+import org.junit.Test;
+
+import javax.money.MonetaryAmount;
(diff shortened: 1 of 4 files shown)


ccsalazarr-commercetools:
@katmatt could you review it? please

commercetools-models/src/test/java/io/sphere/sdk/orders/OrderImportDraftBuilderTest.java
```diff
@@ -0,0 +1,99 @@
...
+import java.time.ZonedDateTime;
+import java.util.List;
+
+import static io.sphere.sdk.test.SphereTestUtils.*;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class OrderImportDraftBuilderTest {
+
+    @Test
+    public void createsFromOrderImportDraft() throws Exception {
```
katmatt: I would name this test like `constructWithAllValuesSet`.

commercetools-models/src/test/java/io/sphere/sdk/orders/OrderImportDraftBuilderTest.java
```diff
@@ -0,0 +1,99 @@
...
+        final LocalizedEnumValue yellow = LocalizedEnumValue.of("yellow", LocalizedString.of(ENGLISH, "yellow").plus(GERMAN, "gelb"));
```
katmatt: We should try to use our fixture classes here.

commercetools-models/src/test/java/io/sphere/sdk/orders/OrderImportDraftBuilderTest.java
```diff
@@ -0,0 +1,99 @@
...
+import org.junit.Test;
+
+import javax.money.MonetaryAmount;
+import java.time.ZonedDateTime;
+import java.util.List;
+
+import static io.sphere.sdk.test.SphereTestUtils.*;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class OrderImportDraftBuilderTest {
```
katmatt: I like that you created a unit test for this builder 👍

commercetools-models/src/test/java/io/sphere/sdk/orders/OrderImportDraftBuilderTest.java
ccsalazarr-commercetools: @katmatt I think that if we use for example `ProductFixtures`, we would have to transform the test into an IntegrationTest, as the already existing methods in the fixture are using the backend to create the products. or what fixtures do you have in mind?

katmatt: You're right and I was wrong. And then I would suggest to keep your code.

PR #1330, opened 2017-03-08
Product search is on stock in channels

### LINKED ISSUE #1312: Add isOnStockInChannels to ProductVariantAvailabilityFilterSearchModel
Add support for https://github.com/sphereio/dev.commercetools.com/pull/448 to our ProductVariantAvailabilityFilterSearchModel.

katmatt:
Closes #1312


commercetools-models/src/main/java/io/sphere/sdk/products/search/ProductVariantAvailabilityFilterSearchModel.java
```diff
@@ -10,6 +11,16 @@
...
 
+    /**
+     * Returns a filter search model for product variant availabilities by channels with
+     * {@link ProductVariantAvailability#isOnStock()} equal to {@code true}.
+     *
+     * {@include.example io.sphere.sdk.products.search.ProductAvailabilitySearchIntegrationTest#searchForIsOnStockInChannels()}
+     *
+     * @return a filter search model for specifying the channels by their id
+     */
+    TermFilterSearchModel<T, String>  isOnStockByChannels();
```
lauraluiz: After some fresh thought now, there's two things I don't like from this name.

The most important is the fact that it shares a name with `isOnStock`, where you define a boolean as input. In our case we have this always as `true` and the input is the channels instead. To express this I would at least suggest removing the `is` part because that denotes a question. I still feel is not expressive enough, but I'm out of better ideas.

Then the second issue is the `by` that I myself suggested. I don't feel it helps in any way now, I would maybe go back to `in`, which at least follows the backend name.

So in short: `onStockInChannels().containsAny(channel1, channel2)`. Not super happy, but well...

Issue #1309, opened 2017-02-09
Implement Product Update Actions "staged" parameter


nkuehn:
To directly publish e.g. the prices that come from an external feed it's necessary to be able to directly publish the update sent. 

This is a supported API feature ( http://dev.commercetools.com/http-api-projects-products.html#set-prices ) but not in the SDK. 

There are a bunch of closed older issues (2015 etc) that mention the topic, but it's not clear why the API feature is not in the SDK.

katmatt:
@nkuehn the current jvm sdk version is lacking the staged update action parameter for all update actions that support it. I would suggest to add it for all update actions and not just the setPrices update action.

nkuehn:
right, I suspected that already but wanted to keep the scope on what I needed atm. 

I suggest you cross-check with the platform team what the background behind the topic is - it's relatively seldom that parameters are "forgotten".  

In fact, there is an old issue about it: https://github.com/commercetools/commercetools-jvm-sdk/issues/598  that implements a "stagedWrapper" in test code. 
https://github.com/commercetools/commercetools-jvm-sdk/blob/69d1bf9fad0aa72cdff6e238355761d2b1142dbc/commercetools-models/src/test/java/io/sphere/sdk/products/commands/ProductUpdateCommandIntegrationTest.java#L947

There may have been design reasons (maybe platform wanted to remove the possibility?), but they are opaque to me. 

@mmoelli do you remember what was the background here?

PR #1334, opened 2017-03-09
Staged parameter product update actions #1309

ccsalazarr-commercetools:
@katmatt could you review this please? ... although it is not ready yet to merge and close the issue, but it would be nice to have a review before continuing with the other update actions

### DIFF
--- a/commercetools-models/src/main/java/io/sphere/sdk/products/commands/updateactions/ChangeAssetName.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/products/commands/updateactions/ChangeAssetName.java
@@ -51,7 +51,7 @@ public final class ChangeAssetName extends StagedProductUpdateActionImpl<Product
     }
 
     public static ChangeAssetName ofVariantId(final Integer variantId, final String assetId, final LocalizedString name) {
-        return ofVariantId(variantId, assetId, name, true);
+        return ofVariantId(variantId, assetId, name, null);
     }
 
     public static ChangeAssetName ofVariantId(final Integer variantId, final String assetId, final LocalizedString name, @Nullable final Boolean staged) {
@@ -59,7 +59,7 @@ public final class ChangeAssetName extends StagedProductUpdateActionImpl<Product
     }
 
     public static ChangeAssetName ofSku(final String sku, final String assetId, final LocalizedString name) {
-        return ofSku(sku, assetId, name, true);
+        return ofSku(sku, assetId, name, null);
     }
 
     public static ChangeAssetName ofSku(final String sku, final String assetId, final LocalizedString name, @Nullable final Boolean staged) {
--- a/commercetools-models/src/main/java/io/sphere/sdk/products/commands/updateactions/ChangeAssetName.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/products/commands/updateactions/ChangeAssetName.java
@@ -1,6 +1,5 @@
 package io.sphere.sdk.products.commands.updateactions;
 
-import io.sphere.sdk.commands.UpdateActionImpl;
 import io.sphere.sdk.models.LocalizedString;
 import io.sphere.sdk.products.Product;
 
@@ -17,7 +16,7 @@ import javax.annotation.Nullable;
  * <p>By SKU (attention, SKU is optional field in a variant):</p>
  * {@include.example io.sphere.sdk.products.commands.ProductUpdateCommandIntegrationTest#changeAssetNameBySku()}
  */
-public final class ChangeAssetName extends UpdateActionImpl<Product> {
+public final class ChangeAssetName extends StagedProductUpdateActionImpl<Product> {
     @Nullable
     private final Integer variantId;
     @Nullable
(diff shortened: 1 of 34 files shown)


commercetools-models/src/test/java/io/sphere/sdk/products/commands/ProductUpdateCommandIntegrationTest.java
```diff
@@ -8,7 +8,10 @@
 import io.sphere.sdk.commands.UpdateAction;
```
ccsalazarr-commercetools: I am not sure now about what is the idea of this `possibleToHackUpdateForStagedAndCurrent` test now, what it does, is to set the staged parameter to false by using the `StagedWrapper`, but now the update actions include the staged parameter.

katmatt: Please remove the `StagedWrapper` and then the `possibleToHackUpdateForStagedAndCurrent`: we don't need them anymore with your fix :wink:

commercetools-models/src/main/java/io/sphere/sdk/products/commands/updateactions/StagedProductUpdateActionImpl.java
```diff
@@ -0,0 +1,21 @@
...
+import io.sphere.sdk.commands.UpdateActionImpl;
+
+import javax.annotation.Nullable;
+
+abstract class StagedProductUpdateActionImpl<T> extends UpdateActionImpl<T> {
+
+    @Nullable
+    protected final Boolean staged;
+
+    protected StagedProductUpdateActionImpl(final String action, final Boolean staged) {
```
katmatt: One last issue: please add `@Nullable` in font of the `final Boolean staged` parameter.

commercetools-models/src/test/java/io/sphere/sdk/products/commands/ProductUpdateCommandIntegrationTest.java
```diff
@@ -944,46 +1297,53 @@ public void changeMasterVariantWithSku() {
...
+                    .as("variant IDs don't change in reordering")
+                    .isNotEqualTo(originalMasterVariantId)
+                    .isEqualTo(variantSupposedToBeMaster.getId());
+
+            return updatedProduct;
+        });
     }
 
     @Test
     public void possibleToHackUpdateForStagedAndCurrent() throws Exception {
```
katmatt: Is this test really still necessary? Does it test a different test case then the test that you created? If it covers a different test case then please rename. Otherwise please just remove this test.

ccsalazarr-commercetools:
@katmatt in the API documentation http://dev.commercetools.com/http-api-projects-products.html#set-asset-sources, it is the setAssetSources product update action, but in the JVM SDK it is not created, should I create it?

commercetools-models/src/main/java/io/sphere/sdk/products/commands/updateactions/ChangeAssetName.java
```diff
@@ -52,10 +51,18 @@ public LocalizedString getName() {
     }
 
     public static ChangeAssetName ofVariantId(final Integer variantId, final String assetId, final LocalizedString name) {
-        return new ChangeAssetName(assetId, variantId, null, name);
+        return ofVariantId(variantId, assetId, name, true);
```
katmatt: We should use `null` instead of `true` here and do this consistently for all update actions. Our API documentation states that the default is `true`, meaning that if we pass in `null` our backend will treat it like `true`. Using `null` here makes it more consistent with our other update actions. And I think if you use the replace from idea like: 
find `, true);` replace `, null);`
and review the changes it shouldn't cause too much work. But please let me know if fixing this issue would require a lot of effort.

commercetools-models/src/test/java/io/sphere/sdk/products/commands/ProductUpdateCommandIntegrationTest.java
```diff
@@ -86,6 +88,42 @@ public void moveImageToPositionByVariantId() throws Exception {
...
             assertThat(urls).containsExactly(url2, url1, url3);
+            assertThat(updatedProduct.getMasterData().hasStagedChanges()).isTrue();
+
+            return updatedProduct;
+        });
+    }
+
+    @Test
+    public void moveImageToPositionByVariantIdWithStaged() {
+        moveImageToPositionByVariantIdWithStaged(true);
```
katmatt: Good idea to extract the tests in a separate method 👍

commercetools-models/src/main/java/io/sphere/sdk/products/commands/updateactions/ChangeAssetName.java
ccsalazarr-commercetools: I will make the change, I assume it is not a big effort

ccsalazarr-commercetools:
@katmatt could you review it again please?

PR #1344, opened 2017-03-16
add deleteDaysAfterLastModification to Cart and CartDraft, close #1337

### LINKED ISSUE #1337: Add TTL field deleteDaysAfterLastModification to Cart
Our cart implementation is currently missing the TTL field `deleteDaysAfterLastModification`. We should add it to `Cart`, `CartDraft` and also add the new `SetDeleteDaysAfterLastModification` [http://dev.commercetools.com/http-api-projects-carts.html#set-deletedaysafterlastmodification-beta](url). And write integration tests for the new fields and the new update action too. 

This changes is very similar to [https://github.com/commercetools/commercetools-jvm-sdk/pull/1336](url).

ccsalazarr-commercetools:
@katmatt could you review it please

### DIFF
--- a/commercetools-models/src/test/java/io/sphere/sdk/carts/commands/CartCreateCommandIntegrationTest.java
+++ b/commercetools-models/src/test/java/io/sphere/sdk/carts/commands/CartCreateCommandIntegrationTest.java
@@ -51,6 +51,19 @@ public class CartCreateCommandIntegrationTest extends IntegrationTest {
     }
 
     @Test
+    public void cartCreationWithTTL() throws Exception {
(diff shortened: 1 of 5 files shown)


commercetools-models/src/test/java/io/sphere/sdk/carts/commands/CartCreateCommandIntegrationTest.java
```diff
@@ -51,6 +51,19 @@ public void execution() throws Exception {
     }
 
     @Test
+    public void cartCreationWithTTL() throws Exception {
```
katmatt: Great that you also created a separate test here 👍

PR #1351, opened 2017-03-23
Code generator primitive update, close #1346

### LINKED ISSUE #1346: Implement a code generator for primitive type update action
We should implement a code generator for update action for resource fields with primitive types (`String`, `Integer`, `Boolean`, ...). The code generator should introduce a new annotation `@HasUpdateActions` which can be added to a resource representation to enable code generation with our annatation processors. A resource annotated with this annotation:
```
@ResourceValue
@HasUpdateAction
public interface MyResource {
   String getName();
   ...
}
```
should then generate an update action like:
```
public final class SetName extends UpdateActionImpl<MyResource> {
  private final String name;

  SetName(final String name) {
    this.name = name;
  }

  public static SetName of(final String name) {
    return new SetName(name);
  }
  ...
}
```
We should use our already existing and [javapoet](https://github.com/square/javapoet) to generate the code.

ccsalazarr-commercetools:
@katmatt I created a pull request in case you want to have a look. We can review it together next week


commercetools-internal-processors/src/main/java/io/sphere/sdk/annotations/processors/HasUpdateActionsProcessor.java
```diff
@@ -0,0 +1,27 @@
...
+import com.squareup.javapoet.JavaFile;
+import io.sphere.sdk.annotations.HasUpdateActions;
+import io.sphere.sdk.annotations.processors.generators.UpdateActionsGenerator;
+
+import javax.annotation.processing.SupportedAnnotationTypes;
+import javax.annotation.processing.SupportedSourceVersion;
+import javax.lang.model.SourceVersion;
+import javax.lang.model.element.TypeElement;
+import java.util.List;
+
```
katmatt: It would be nice if you would add some javadoc here, like we did for the other generators.

commercetools-internal-processors/src/main/java/io/sphere/sdk/annotations/processors/generators/AbstractMultipleGenerator.java
```diff
@@ -0,0 +1,30 @@
...
+
+import com.squareup.javapoet.JavaFile;
+import com.squareup.javapoet.TypeSpec;
+
+import javax.lang.model.element.TypeElement;
+import javax.lang.model.util.Elements;
+import java.util.ArrayList;
+import java.util.List;
+
+abstract class AbstractMultipleGenerator extends BaseAbstractGenerator {
```
katmatt: What about changing the name of this class to sth like `AbstractMultipleFilegenerator`? And adding a short javadoc with an explanation would be nice too 😉

commercetools-internal-processors/src/main/java/io/sphere/sdk/annotations/processors/generators/AbstractMultipleGenerator.java
```diff
@@ -0,0 +1,30 @@
...
+
+    AbstractMultipleGenerator(final Elements elements) {
+        super(elements);
+    }
+
+    public final List<JavaFile> generate(final TypeElement annotatedTypeElement) {
+        List<JavaFile> javaFiles = new ArrayList<>();
+        final List<TypeSpec> typeSpecs = generateTypes(annotatedTypeElement);
+        typeSpecs.forEach(typeSpec -> {
```
katmatt: Here I personally would use a `for` loop, but it's a matter of taste and we can keep the foreach here.

commercetools-internal-processors/src/main/java/io/sphere/sdk/annotations/processors/generators/BaseAbstractGenerator.java
```diff
@@ -0,0 +1,112 @@
...
+     * @return methods sorted by their {@link PropertyGenModel#getPropertyName(ExecutableElement)}
+     */
+    protected List<ExecutableElement> getAllPropertyMethodsSorted(TypeElement typeElement) {
+        return typeUtils.getAllPropertyMethods(typeElement)
+                .sorted(Comparator.comparing(methodName -> escapeJavaKeyword(PropertyGenModel.getPropertyName(methodName))))
+                .collect(Collectors.toList());
+    }
+
+    /**
+     * Returns all property methods - sorted by their property name.
```
katmatt: We just copied the javadoc from the original method, but we should update the javadoc because this method only returns the direct properties and not `all`.

commercetools-internal-processors/src/main/java/io/sphere/sdk/annotations/processors/generators/UpdateActionsGenerator.java
```diff
@@ -0,0 +1,107 @@
...
+        return super.getPackageName(annotatedTypeElement).concat(".commands.updateactions");
+    }
+
+    @Override
+    public List<TypeSpec> generateTypes(final TypeElement annotatedTypeElement) {
+        final List<ExecutableElement> propertyMethods = getPropertyMethodsSorted(annotatedTypeElement);
+        final List<TypeSpec> typeSpecList = propertyMethods.stream()
+                .filter(m -> isPrimitiveType(PropertyGenModel.of(m).getType()))
+                .map(propertyMethod -> {
+                    final PropertyGenModel property = PropertyGenModel.of(propertyMethod);
```
katmatt: How about extracting this lambda into a separate method? Then we could replace the lambda with a method handle to the extracted method. Let's discuss this separately.

commercetools-internal-processors/src/main/java/io/sphere/sdk/annotations/processors/generators/UpdateActionsGenerator.java
```diff
@@ -0,0 +1,107 @@
...
+                    if (property.isOptional()) {
+                        typeSpecBuilder.addMethod(createUnsetMethod(updateActionClassName));
+                    }
+                    return typeSpecBuilder.build();
+                }).collect(Collectors.toList());
+
+        return typeSpecList;
+    }
+
+    protected boolean isPrimitiveType(final TypeName propertyType) {
```
katmatt: Cood idea to extract this into a method 👍 If we in the future need this method in other classes, we can still move it to our `TypeUtils` class.

commercetools-internal-processors/src/main/java/io/sphere/sdk/annotations/processors/generators/UpdateActionsGenerator.java
```diff
@@ -0,0 +1,107 @@
...
+        return propertyType.isPrimitive() || propertyType.isBoxedPrimitive() || propertyType.toString().equals("java.lang.String");
+    }
+
+    private MethodSpec createUnsetMethod(final String updateActionClassName) {
```
katmatt: Good idea to separate this into a method 👍

commercetools-internal-processors/src/main/java/io/sphere/sdk/annotations/processors/models/TypeUtils.java
```diff
@@ -89,6 +89,23 @@ public ClassName getBuilderType(final TypeElement typeElement) {
                 .filter(this::isPropertyMethod);
     }
 
+
+    /**
+     * Returns all property methods as stream.
```
katmatt: Here we should also update the copied javadoc - because this method only returns the direct properties of the given `TypeElement`. Additionally you should make the parameter `final` and check the original method too - looks like I forgot to add the `final` 😉

commercetools-internal-processors/src/test/java/io/sphere/sdk/annotations/processors/generators/AbstractMultipleGeneratorTest.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.List;
+
+public abstract class AbstractMultipleGeneratorTest {
+
+    @Rule
+    public CompilationRule compilationRule = new CompilationRule();
+
+    protected AbstractMultipleGenerator generator;
+
+    protected List<String> expectedContent(final Class<?> clazz) throws Exception {
```
katmatt: Very good idea to return a list of the expected content here 👍 It would be even nicer if we could add a method to our `AbstractMultipleGenerator` which just returns the `ClassName`s that it will generate for a given `TypeElement`. This would then avoid that we have to generate the actual `TypeSpec`s here.

commercetools-internal-processors/src/test/java/io/sphere/sdk/annotations/processors/generators/AbstractMultipleGeneratorTest.java
```diff
@@ -0,0 +1,56 @@
...
+            try {
+                expectedContentList.add(IOUtils.toString(resourceAsStream, Charsets.UTF_8));
+            } catch (IOException e) {
+                throw new IllegalStateException("Fixture file '" + fixtureFile + "' not found.", e);
+            }
+        });
+        return expectedContentList;
+    }
+
+    protected List<String> generateAsStrings(final Class<?> clazz) throws Exception {
```
katmatt: Very good idea to generate the content for all types 👍

commercetools-internal-processors/src/test/java/io/sphere/sdk/annotations/processors/generators/UpdateActionsGeneratorTest.java
```diff
@@ -0,0 +1,28 @@
...
+    public void setup() {
+        generator = new UpdateActionsGenerator(compilationRule.getElements());
+    }
+
+    @Test
+    public void createMultipleUpdateActions() throws Exception {
+        final List<String> contentList = generateAsStrings(MyResource.class);
+        final List<String> expectedContent = expectedContent(MyResource.class);
+        int index = 0;
+        for (String content : contentList) {
```
katmatt: How about using a for loop here `for (int index = 0; index < contentList.size; i++) {...`?

commercetools-internal-processors/src/test/java/io/sphere/sdk/annotations/processors/generators/UpdateActionsGeneratorTest.java
```diff
@@ -0,0 +1,28 @@
...
+    @Before
+    public void setup() {
+        generator = new UpdateActionsGenerator(compilationRule.getElements());
+    }
+
+    @Test
+    public void createMultipleUpdateActions() throws Exception {
+        final List<String> contentList = generateAsStrings(MyResource.class);
+        final List<String> expectedContent = expectedContent(MyResource.class);
+        int index = 0;
```
katmatt: I would add an `assertThat(contentlist.size()).isEquaTo(expectedContent.size())` here. This makes it clear what we expect.

commercetools-internal-processors/src/test/resources/io/sphere/sdk/annotations/processors/generators/examples/ChangeQuantity.java.expected
```diff
@@ -0,0 +1,32 @@
...
+  private ChangeQuantity(final Integer quantity) {
+    super("changeQuantity");
+    this.quantity = quantity;
+  }
+
+  public Integer getQuantity() {
+    return quantity;
+  }
+
+  public static ChangeQuantity of(final Integer quantity) {
```
katmatt: Here it would be nice to add some javadoc which links to the corresponding get method of the resource:
```
 * @param quantity the quantity {@link MyResource#getquantity()}
```
This will give our users more context how this update action relates to the resource.

commercetools-sdk-base/src/main/java/io/sphere/sdk/annotations/HasUpdateActions.java
```diff
@@ -0,0 +1,4 @@
+package io.sphere.sdk.annotations;
+
+public @interface HasUpdateActions {
```
katmatt: Please add some javadoc to this annotation, that describes that this annotation currently only enables the generation of update actions for primitive typed properties.

ccsalazarr-commercetools:
@katmatt could you review it again, please?

PR #1353, opened 2017-03-27
Tiered prices

### LINKED ISSUE #1348: Add support for tiered prices

katmatt:
Closes #1348


commercetools-models/src/main/java/io/sphere/sdk/products/PriceDraftBuilder.java
```diff
@@ -93,6 +96,11 @@ public PriceDraftBuilder value(final MonetaryAmount value) {
         return this;
     }
 
+    public PriceDraftBuilder tiers(final List<PriceTier> tiers) {
```
ccsalazarr-commercetools: here do we need the `@Nullable` in the method parameter?

PR #1362, opened 2017-03-30
Tiered prices query model

### LINKED ISSUE #1361: Add query model for tiered prices
Our last version didn't had support for querying tiered prices and we should add support for it in our next release.

katmatt:
Closes #1361


commercetools-models/src/test/java/io/sphere/sdk/products/queries/ProductQueryIntegrationTest.java
```diff
@@ -169,6 +167,30 @@ public void queryByReviewRating() {
...
+    public void queryByTiersWithValue() {
+        withProduct(client(), product -> {
+            final ProductQuery productQuery = ProductQuery.of()
+                    .withPredicates(m -> m.masterData().current().variants().prices().tiers().value().currencyCode().is("EUR"))
+                    .plusPredicates(m -> m.is(product));
+
+            final List<Product> results = client().executeBlocking(productQuery).getResults();
+            assertThat(results).hasSize(0);
+        });
+    }
```
lauraluiz: I guess for query just demonstrating that this doesn't fail and it filters the product out is enough. In search endpoint this wouldn't be enough because entering an invalid path/value doesn't fail and filters out everything, so the assertion must be positive (finding something). So it's fine :)

PR #1374, opened 2017-04-26
Add subscriptions

### LINKED ISSUE #1338: Add support for Subscriptions
Add support for [http://dev.commercetools.com/http-api-projects-subscriptions.html](url) to our SDK. Biggest issue is to have some integration tests. I will first start without providing correct MQ destinations and just test that our resource representations are correct. Afterwards I will try to write an integration test with IronMQ or AWS. The biggest challenge here is to integrate it into our travis build.

katmatt:
Closes #1338


commercetools-models/src/main/java/io/sphere/sdk/subscriptions/MessageSubscription.java
```diff
@@ -0,0 +1,21 @@
...
+
+@JsonDeserialize(as = MessageSubscriptionImpl.class)
+@ResourceValue
+public interface MessageSubscription {
+    String getResourceTypeId();
+
+    @Nullable
+    List<String> getTypes();
+
+    static MessageSubscription of(final String resourceTypeId, final String... types) {
```
lauraluiz: I would avoid using varargs or I would have it as syntactic sugar for `of(final String resourceTypeId, final List<String> types)`. Having it as the only method forces developers to use ugly arrays if they are providing types from a list.

commercetools-models/src/test/java/io/sphere/sdk/subscriptions/ChangeSubscriptionTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+        assertThat(jsonNode.get("resourceTypeId").asText()).isEqualTo(Review.resourceTypeId());
+    }
+
+    @Test
+    public void deserialize() throws Exception {
+        ChangeSubscription changeSubscription = SphereJsonUtils.readObject("{\"resourceTypeId\":\"category\"}", ChangeSubscription.class);
+        assertThat(changeSubscription.getResourceTypeId()).isEqualTo(Category.resourceTypeId());
+    }
+}
```
lauraluiz: I like UT! :D

commercetools-models/src/test/java/io/sphere/sdk/subscriptions/SnsUtils.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public class SnsUtils {
+    private SnsUtils() {
+    }
+
+    /**
+     * Creates a new sns test topic.
+     *
+     * @param snsClient the sqs client
+     * @return the arn of the created sns topic
```
lauraluiz: It took me a while to realize that ARN was an acronym for some AWS stuff. Could we write it uppercase in the Javadoc to leave it clear it's not a word? Same with SQS and SNS.

yanns: +1

katmatt: That's a good idea 👍 For me it was just another crazy acronym ;-)

commercetools-models/src/main/java/io/sphere/sdk/subscriptions/MessageSubscription.java
katmatt: Good point and I agree with you, it's better to use lists here 😄

PR #1394, opened 2017-05-15
Adding code generation for ProductDraftFactory and ProductDraftImpl. …

acbeni:
…removing manually generated classes.


commercetools-models/src/main/java/io/sphere/sdk/products/ProductDraftBuilder.java
```diff
@@ -67,13 +35,9 @@ public static ProductDraftBuilder of(final ResourceIdentifiable<ProductType> pro
...
-        return new ProductDraftBuilder(productType.toResourceIdentifier(), name, slug, masterVariant);
+        return of(productType.toResourceIdentifier(),name,slug,masterVariant);
     }
 
-    @Override
-    public ProductDraft build() {
-        return new ProductDraftImpl(productType, getName(), getSlug(), getDescription(), getCategories(), getMetaTitle(), getMetaDescription(), getMetaKeywords(), masterVariant, variants, getTaxCategory(), getSearchKeywords(), getState(), getCategoryOrderHints(), publish, key);
-    }
 
     protected ProductDraftBuilder getThis() {
```
katmatt: I don't know why we ever had this method and would suggest to remove/inline it.

commercetools-models/src/main/java/io/sphere/sdk/products/ProductDraftBuilder.java
```diff
@@ -158,66 +71,13 @@ public ProductDraftBuilder categories(final List<Reference<Category>> categories
...
-
-    public SearchKeywords getSearchKeywords() {
-        return searchKeywords;
-    }
-
-    public Reference<TaxCategory> getTaxCategory() {
-        return taxCategory;
-    }
 
     public ProductDraftBuilder taxCategory(final Referenceable<TaxCategory> taxCategory) {
```
katmatt: This method and the `state`are already generated, so we should remove them here.

commercetools-models/src/main/java/io/sphere/sdk/products/ProductDraftBuilder.java
```diff
@@ -158,66 +71,13 @@ public ProductDraftBuilder categories(final List<Reference<Category>> categories
      */
     public ProductDraftBuilder categoriesAsObjectList(final List<Category> categories) {
         final List<Reference<Category>> referenceList = categories.stream()
-                .filter(x -> x != null)
-                .map(cat -> cat.toReference())
+                .filter(Objects::nonNull)
```
katmatt: Good idea to replace the lambdas with method handles 👍

commercetools-models/src/main/java/io/sphere/sdk/products/ProductDraft.java
```diff
@@ -62,12 +69,13 @@
...
     /**
      * Flag for publishing the product immediately.
-     *
+     * <p>
      * {@include.example io.sphere.sdk.products.commands.ProductCreateCommandIntegrationTest#createPublishedProduct()}
      *
      * @return true if product should be published
      */
     @Nullable
+    @JsonProperty
```
katmatt: Just for consistency with our code and to make the name of this property more explicit, I would suggest to add the value "publish" to this annotation: `@JsonProperty("publish")`

PR #1417, opened 2017-05-30
creation of getter for sphere config/ project key through SphereClient.

### LINKED ISSUE #1403: Getter for sphere config/ project key through SphereClient
### Description

It would be great to have on the `SphereClient` a getter for the the `SphereApiConfig`, or even the projectKey directly, to be able to get the project key easily just through the client without needing to have an instance of `SphereApiConfig` beforehand.

### Expected Behaviour
Ability to get the project key from a client like that:
````java
final SphereClient client = SphereClient.of(clientConfig, httpClient, tokenSupplier);
String projectKey = client.getConfig().getProjectKey();
````

or even better:

````java
final SphereClient client = SphereClient.of(clientConfig, httpClient, tokenSupplier);
String projectKey = client.getProjectKey();
````

### Current Behaviour
Currently it's not possible to access the config/project key from just the client.

### Context
In the [commercetools sync](https://github.com/commercetools/commercetools-sync-java/) library, a user of the library is expected to only pass a `sphereClient` to the library. So it doesn't actually know about the Config. However, the library needs to add the project key, when logging exceptions to the user for easier debugging. Therefore, with the current behaviour of the JVM SDK, the user would need to pass the config as well so the library can have this project key.

### PR DESCRIPTION
**Pull request by acbeni**
Closes #1403

### DIFF
--- a/commercetools-java-client-core/src/main/java/io/sphere/sdk/client/BlockingSphereClientImpl.java
+++ b/commercetools-java-client-core/src/main/java/io/sphere/sdk/client/BlockingSphereClientImpl.java
@@ -45,4 +45,9 @@ final class BlockingSphereClientImpl extends Base implements BlockingSphereClien
     public <T> T executeBlocking(final SphereRequest<T> sphereRequest, final Duration duration) {
         return executeBlocking(sphereRequest, duration.toMillis(), TimeUnit.MILLISECONDS);
     }
+
+    @Override
+    public SphereApiConfig getConfig() {
+        return null;
+    }
 }
--- a/commercetools-java-client-core/src/main/java/io/sphere/sdk/client/BlockingSphereClientImpl.java
+++ b/commercetools-java-client-core/src/main/java/io/sphere/sdk/client/BlockingSphereClientImpl.java
@@ -48,6 +48,6 @@ final class BlockingSphereClientImpl extends Base implements BlockingSphereClien
 
     @Override
     public SphereApiConfig getConfig() {
-        return null;
+        return this.delegate.getConfig();
     }
 }
--- a/commercetools-java-client-core/src/main/java/io/sphere/sdk/client/TestDoubleSphereClientFactory.java
+++ b/commercetools-java-client-core/src/main/java/io/sphere/sdk/client/TestDoubleSphereClientFactory.java
@@ -16,6 +16,12 @@ import static io.sphere.sdk.utils.CompletableFutureUtils.successful;
 
  */
 public final class TestDoubleSphereClientFactory extends Base {
+
+    /**
+     * Dummy configuraion for test clients
+     */
+    private static final SphereApiConfig sphereApiConfig = SphereApiConfig.of("fake-project-key-for-testing", "https://createHttpTestDouble.tld");
+
     private TestDoubleSphereClientFactory() {
     }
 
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

**Comment by lauraluiz** [comment]
@acbeni let me know when can I review it (in case you are still working on it) :)

**Comment by acbeni** [comment]
@lauraluiz Hi i am done for this issue 👍  please feel free to review it.

File: commercetools-java-client-core/src/main/java/io/sphere/sdk/client/TestDoubleSphereClientFactory.java [hunk 1]
```diff
@@ -81,6 +86,11 @@ public void close() {
             public String toString() {
                 return "SphereClientObjectTestDouble";
             }
+
+            @Override
+            public SphereApiConfig getConfig() {
+                return null;
```
lauraluiz: I don't feel it right that in a class delivered by our library we have a method not signalled with `@Nullable` that returns a null. Of course we don't have anything meaningful to return because it's a fake client...
But checking the code I see that in one of the methods there is one SphereApiConfig with fake data. Maybe we can go on that line and provide a dummy config with a project key and API URL that leaves no doubt this is a fake config. Maybe @katmatt has a better idea?

katmatt:
@lauraluiz @acbeni I just realized that our `SphereClientConfig` class has public `of` factory methods, that should make it quite easy to return a client config instead of `null`. And we could even initialize the fields with fake values. What do you think about it?

lauraluiz:
Do you mean SphereClientConfig or SphereApiConfig? We only need the latter.

katmatt:
@lauraluiz Sorry for the confusion, I was talking about `SphereApiConfig`.

lauraluiz:
Okay then yes, that was suggested approach as well 👍 
I would probably skip it for tests though, since that is under our control and provides no benefits.

acbeni:
@katmatt @lauraluiz Yep! i agree with using a dummy config, actually didn't found something meaningful to put as a return so i used null. still it's a good idea to eliminate the null on a non nullable getter.

commercetools-java-client-core/src/main/java/io/sphere/sdk/client/BlockingSphereClientImpl.java
```diff
@@ -45,4 +45,9 @@ public void close() {
     public <T> T executeBlocking(final SphereRequest<T> sphereRequest, final Duration duration) {
         return executeBlocking(sphereRequest, duration.toMillis(), TimeUnit.MILLISECONDS);
     }
+
+    @Override
+    public SphereApiConfig getConfig() {
+        return null;
```
acbeni: Yep, sorry my bad :(

commercetools-java-client-core/src/main/java/io/sphere/sdk/client/TestDoubleSphereClientFactory.java
```diff
@@ -55,7 +55,9 @@ public String toString() {
 
             @Override
             public SphereApiConfig getConfig() {
-                return null;
+
+                return SphereApiConfig.of("DUMMY_PROJECT_KEY","DUMMY_API_URL");
```
lauraluiz: Notice there's another `getConfig` some lines below this code.
Also notice that this client's `execute` method already uses a `SphereApiConfig`. Why don't we better save the config as a class field and use it in both sides?

PR #1445, opened 2017-06-21
Use type.getId() instead of type.getTypeId() in the builder, close #1444

### LINKED ISSUE #1444: CustomFields copy factory methods copies the `TypeId` instead of the `id` of the Type
### Description
There is a bug in the copy factory method of custom fields. It happens when building a `CutomFieldsDraft` instance from a `CustomFields` instance. The `id` is confused with the `TypeId`. 

I noticed it when converting a list of `Category`s to `CategoryDrafts` with custom fields. I got the following response from the platform:
```json
http response formatted body: {
  "statusCode" : 400,
  "message" : "Request body does not contain valid JSON.",
  "errors" : [ {
    "code" : "InvalidJsonInput",
    "message" : "Request body does not contain valid JSON.",
    "detailedErrorMessage" : "actions -> type -> id: Invalid UUID: 'type'"
  } ]
}
````
### The problem
The problem is here in this [line](https://github.com/commercetools/commercetools-jvm-sdk/blob/master/commercetools-models/src/main/java/io/sphere/sdk/types/CustomFieldsDraftBuilder.java#L58):
````java
public static CustomFieldsDraftBuilder of(final CustomFields customFields) {
        final Reference<Type> type = customFields.getType();
        return new CustomFieldsDraftBuilder(type.getTypeId(), type.getKey()).fields(customFields.getFieldsJsonMap());
    }
```` 
This causes the Custom Type to have the id "type" (which is the `typeId`) instead of an actual `id`.

### The fix
````java
public static CustomFieldsDraftBuilder of(final CustomFields customFields) {
        final Reference<Type> type = customFields.getType();
        return new CustomFieldsDraftBuilder(type.getId(), type.getKey()).fields(customFields.getFieldsJsonMap());
    }
````

### PR DESCRIPTION
**Pull request by ccsalazarr-commercetools**
@katmatt could you review it please

### DIFF
--- a/commercetools-models/src/test/java/io/sphere/sdk/types/CustomFieldsDraftBuilderIntegrationTest.java
+++ b/commercetools-models/src/test/java/io/sphere/sdk/types/CustomFieldsDraftBuilderIntegrationTest.java
@@ -1,0 +1,31 @@
+package io.sphere.sdk.types;
+
+import io.sphere.sdk.categories.CategoryDraftBuilder;
+import io.sphere.sdk.test.IntegrationTest;
+import org.junit.Test;
+
+import static io.sphere.sdk.categories.CategoryFixtures.withCategory;
+import static io.sphere.sdk.test.SphereTestUtils.en;
+import static io.sphere.sdk.types.TypeFixtures.STRING_FIELD_NAME;
+import static io.sphere.sdk.types.TypeFixtures.withUpdateableType;
+import static java.util.Collections.singletonMap;
+import static org.assertj.core.api.AssertionsForInterfaceTypes.assertThat;
+
+public class CustomFieldsDraftBuilderIntegrationTest extends IntegrationTest {
+
+    @Test
+    public void copyFromCustomField() {
+        withUpdateableType(client(), type -> {
+            final String value = "foo";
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: commercetools-models/src/test/java/io/sphere/sdk/types/CustomFieldsDraftBuilderIntegrationTest.java [hunk 1]
```diff
@@ -0,0 +1,31 @@
...
+import org.junit.Test;
+
+import static io.sphere.sdk.categories.CategoryFixtures.withCategory;
+import static io.sphere.sdk.test.SphereTestUtils.en;
+import static io.sphere.sdk.types.TypeFixtures.STRING_FIELD_NAME;
+import static io.sphere.sdk.types.TypeFixtures.withUpdateableType;
+import static java.util.Collections.singletonMap;
+import static org.assertj.core.api.AssertionsForInterfaceTypes.assertThat;
+
+public class CustomFieldsDraftBuilderIntegrationTest extends IntegrationTest {
```
katmatt: Great that you added an unit test 👍

katmatt:
Closes #1445

PR #1452, opened 2017-06-26
Add missing update action for attribute definition, close #1435

### LINKED ISSUE #1435: Add missing update actions for AttributeDefinition
Our SDK currently doesn't support all update actions for the ÀttributeDefinition` type and we should add them to our SDK. Since noone requested them, this issue isn't very urgent and is more about the completeness of our SDK. 

But may be we can add them with the next major update of our SDK.

ccsalazarr-commercetools:
@katmatt could you review it please.


commercetools-models/src/main/java/io/sphere/sdk/types/commands/updateactions/ChangeKey.java
```diff
@@ -1,29 +0,0 @@
-package io.sphere.sdk.types.commands.updateactions;
```
katmatt: Nice that you replaced this update action with a generated one 👍

commercetools-models/src/main/java/io/sphere/sdk/producttypes/commands/updateactions/SetInputTip.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * Sets the attribute input tip. Allows to set additional information about the specified attribute.
+ *
+ * {@doc.gen intro}
+ *
+ * {@include.example io.sphere.sdk.producttypes.commands.ProductTypeUpdateCommandIntegrationTest#setInputTip()}
+ *
+ * @see AttributeDefinition#getInputTip()
```
katmatt: Adding a link helps to understand the relation between the resource and this update action 👍

Issue #1434, opened 2017-06-15
Add getPage method to PagedResult interface


katmatt:
We should add a `int getPage()` method to our `PagedResult` interface, which calculates the page number of the result. 

We should add it as a default method, so that we don't break the binary compatibility of our SDK.

lauraluiz:
Sorry for the offtopic, but do we have an issue for the total pages too? That's the most complex to calculate :)

katmatt:
@lauraluiz Thanks for raising this 👍  And it's not off topic, I just forgot to add it to our issue 😞 I created a new one 😉 
https://github.com/commercetools/commercetools-jvm-sdk/issues/1470

acbeni:
@lauraluiz @katmatt , it has been taken care of the `getPagesCount()` and has been added to the pull request.

PR #1460, opened 2017-07-03
Add get page method to paged result interface

acbeni:


commercetools-sdk-base/src/main/java/io/sphere/sdk/queries/PagedResult.java
```diff
@@ -50,6 +50,22 @@
      }
 
     /**
+     * calculates the page number of the result.
```
katmatt: Great that you added javadoc, but the sentence should start with a capital letter 😉

commercetools-sdk-base/src/main/java/io/sphere/sdk/queries/PagedResult.java
```diff
@@ -50,6 +50,22 @@
...
+     *
+     * @return the page number of the result.
+     */
+    default long getPage() {
+
+        if (getOffset() == null) {
```
katmatt: I think I understand your intention, but I'm not sure if throwing an ÙnsupportedOperationException` is a good idea here. We already use `Long` everywhere in this class and they aren't nullable. I would suggest to use `Long` here two.

commercetools-sdk-base/src/main/java/io/sphere/sdk/queries/PagedResult.java
```diff
@@ -50,6 +50,22 @@
...
+            throw new UnsupportedOperationException("Can only be used if the offset is known.");
+        }
+        if (getCount() == null || getCount() <= 0) {
```
katmatt: And here it's save to assume that `getCount()` isn't null and that it's always positive.

commercetools-sdk-base/src/main/java/io/sphere/sdk/queries/PagedResult.java
acbeni: I was actually following the same design pattern used on the class, to make it easy to dete defectuous implementations.
But as you said lets remove the check since the field is not nullable. 👍

commercetools-sdk-base/src/main/java/io/sphere/sdk/queries/PagedResult.java
acbeni: 👍 I am ok with assumption that the null check should be removed, still we should keep the check that getCount() is strictly positive since you can request paged query with 0 as a size.
in that case the page concept doesn't make sense.

commercetools-sdk-base/src/main/java/io/sphere/sdk/queries/PagedResult.java
acbeni: Good catch 👍

katmatt:
Closes #1434

commercetools-sdk-base/src/main/java/io/sphere/sdk/queries/PagedResult.java
```diff
@@ -50,6 +50,20 @@
...
 
     /**
+     * Calculates the page number of the result.
+     *
+     * @return the page number of the result.
+     */
+    default Long getPage() {
+
+        if (getCount() == 0) {
+            throw new UnsupportedOperationException("Can only be used if the count is known, and is greater than 0");
```
katmatt: I would prefer that we don't throw an exception here and instead just return `0L`. Otherwise a user of this method needs to add a `try/catch` block. Your solution is correct in a mathematical sense, but as a user of our SDK I don't like to catch unchecked exceptions.

acbeni: Nice catch 😉 

Now done 👍

commercetools-sdk-base/src/test/java/io/sphere/sdk/queries/PagedQueryResultTest.java
```diff
@@ -44,6 +46,32 @@ public void middleOfManyResult() throws Exception {
     }
 
     @Test
+    public void getResultPage() throws Exception {
+        final PagedQueryResult<Integer> queryResult = PagedQueryResult.of((long) 80, TOTAL, listOfSize(PAGE_SIZE));
+        assertThat(queryResult.getPage()).isEqualTo(3);
```
lauraluiz: The data is a bit unrealistic so it's difficult to know the page it should be. If you had offset 80 and total products of 100, then you should not be receiving 25 products, as 80 is not divisible by 25. So it's difficult to determine if the test is good.
So let's say that you are getting 25 products back because you are neither at the 1st or the last page. Then we could say we requested an offset of 75 for example, in which case the page should be 3 (starting from 0).
In general I would recommend to find edge cases, not based on your implementation, but on the problem you are solving. With pages that could be:
a) request didn't raise any results at all (offset = 0, total = 0, count = 0, page = 0)
b) request has more than one page and ends with a count < page size:
   1. request first page (offset = 0, total = 106, count = 25, page = 0)
   2. request middle page (offset = 75, total = 106, count = 25, page = 3)
   3. request last page (offset = 100, total = 106, count = 6, page = 4)
   4. request past the last page (offset = 125, total = 106, count = 0, page = ?? as we don't know the requested page size we cannot know the page, but returning 0 can be misleading...)

Pagination might look like easy to implement, but it's very tricky. We should better have strong tests behind it :)
And I guess we should assume the backend doesn't return inconsistent results, but it's worth asking them if that can happen. With facets that could happen, for example.

commercetools-sdk-base/src/main/java/io/sphere/sdk/queries/PagedResult.java
```diff
@@ -50,6 +50,32 @@
...
+     * Calculates the total number of pages matching the request.
+     *
+     * @return the total number of pages , 0 if the page size is 0, meaning "{@code getCount()}" returns 0.
+     */
+    default Long getPagesCount() {
+
+        if(getCount() == 0)
+            return 0L;
+
+        return   getTotal() / getCount();
```
lauraluiz: My implementation looks a bit different from it:
```
        if (pagedResult.isLast()) {
            return currentPage;
        } else if (pagedResult.getCount() > 0) {
            final Double totalPages = Math.ceil(pagedResult.getTotal() / pagedResult.getCount());
            return totalPages.longValue();
        } else {
            return 0;
        }
```
But let's check after you write more tests.

commercetools-sdk-base/src/test/java/io/sphere/sdk/queries/PagedQueryResultTest.java
```diff
@@ -44,6 +46,19 @@ public void middleOfManyResult() throws Exception {
     }
 
     @Test
+    public void getResultPage() throws Exception {
+        final PagedQueryResult<Integer> queryResult = PagedQueryResult.of((long) 80, TOTAL, listOfSize(PAGE_SIZE));
```
acbeni: 👍

commercetools-sdk-base/src/main/java/io/sphere/sdk/queries/PagedResult.java
```diff
@@ -50,6 +50,32 @@
...
+
+        return getOffset() / getCount();
+    }
+
+    /**
+     * Calculates the total number of pages matching the request.
+     *
+     * @return the total number of pages , 0 if the page size is 0, meaning "{@code getCount()}" returns 0.
+     */
+    default Long getPagesCount() {
```
acbeni: I also was suffering figuring out a name, but lets consider `getTotalPages` since it seem pretty clear.

commercetools-sdk-base/src/test/java/io/sphere/sdk/queries/PagedQueryResultTest.java
acbeni: 👍 very good catch actually ny algo was bugus i guess, 
for the remarque: 
>request past the last page (offset = 125, total = 106, count = 0, page = ??
i think the best solution is to throw an 'IllegalStateException' with message 'Page number greater then total number of pages' and then even if its a misbehaviour from the backend it should be raised to be corrected.
What do you think?

acbeni: or maybe just show the total page number which is the max, since it would be inconvenient to block the whole page from loading because of a small bug form the backend
i think that idea is much better.

commercetools-sdk-base/src/main/java/io/sphere/sdk/queries/PagedResult.java
```diff
@@ -50,6 +50,32 @@
      }
 
     /**
+     * Calculates the page number of the result.
```
acbeni: @lauraluiz good catch
Change done!

commercetools-sdk-base/src/main/java/io/sphere/sdk/queries/PagedResult.java
```diff
@@ -50,8 +50,8 @@
      }
 
     /**
-     * Calculates the page number of the result, the page numbers starts with 0, so if you need to use it in the UI, starting from page 1 you need to use
-     *{@code getPageIndex() + 1}
+     * Calculates the page number of the result, the pages are indexed staring 0, which means that {@code getPageIndex()}
+     * returns a value in [0,n) , given 'n' is the total number of pages
```
lauraluiz: 👍

acbeni:
@lauraluiz now all the tests are fine so if its ok for you, maybe we can merge this branche
what do you think?

commercetools-sdk-base/src/test/java/io/sphere/sdk/queries/PagedQueryResultTest.java
```diff
@@ -38,12 +40,47 @@ public void firstOfManyResult() throws Exception {
...
+        }
+    }
+
+
+
+
+    @Test
+    public void getPagesCountForEmptyResults() throws Exception {
+        final PagedQueryResult<Integer> queryResult = PagedQueryResult.of( 80L, TOTAL, listOfSize(0));
+        assertThat(queryResult.getTotalPages()).isEqualTo(0);
```
lauraluiz: So in here you are saying there are no pages when actually there are, we just can't say how many because in the response we have no information about the limit (a pity, actually, but nothing we can do about it).
But just because we can't know, we should not give a number that would be valid in another situation. That's why I suggested in another comment (coming from a conversation with @katmatt ) to return an `Optional` instead of a number, and leave it clear the situations in which we can't tell. This would apply for the page index method too.

commercetools-sdk-base/src/test/java/io/sphere/sdk/queries/PagedQueryResultTest.java
```diff
@@ -38,12 +40,47 @@ public void firstOfManyResult() throws Exception {
...
+            assertThat(queryResult.getTotalPages()).isEqualTo(5);
+        }
+        {
+            final PagedQueryResult<Integer> queryResult = PagedQueryResult.of(75L, 106l, listOfSize(25));
+            assertThat(queryResult.getPageIndex()).isEqualTo(3);
+            assertThat(queryResult.getTotalPages()).isEqualTo(5);
+        }
+
+        {
+            final PagedQueryResult<Integer> queryResult = PagedQueryResult.of(100L, 106l, listOfSize(25));
```
lauraluiz: This is not a valid answer, because if you are putting an offset of 100 and you have a total of 106 products, then the size of the results you are going to receive is 6. Then with the current implementation you would receive 0, which is wrong. Applying the solution I proposed in the other comment would not return anything (absent) and it would be one of those situations where we can't tell the page structure.

katmatt:
@lauraluiz @acbeni After reading @lauraluiz comment that the `limit` isn't part of the http response from our API, I was wondering how difficult it actually is to copy the `limit` request parameter from our query object to the paged result object?

lauraluiz:
Yep, that would be another way, I had completely forgotten that evident option :P
Only question would be if that affects badly our or other clients' architecture. I always wondered why we receive the offset but not the limit though.

commercetools-sdk-base/src/main/java/io/sphere/sdk/queries/PagedResultBase.java
```diff
@@ -6,12 +6,14 @@
 
 public abstract class PagedResultBase<T> extends Base implements PagedResult<T> {
     protected final Long offset;
+    protected final Long limit;
     protected final Long total;
     protected final List<T> results;
     protected final Long count;
 
-    protected PagedResultBase(final Long offset, final Long total, final List<T> results, final Long count) {
+    protected PagedResultBase(final Long offset, final Long limit, final Long total, final List<T> results, final Long count) {
```
lauraluiz: So I have added everywhere the limit next to the offset because they belong together, but on the other side it might be error-prone for people extending this class as it's a list of Longs and limit is in the middle. You mistake the position and you got it wrong. WDYT?

lauraluiz:
@acbeni @katmatt please review with the limit on the result 🎉

commercetools-sdk-base/src/main/java/io/sphere/sdk/queries/PagedResultBase.java
katmatt: I think that it's ok to have this somehow ugly constructor. The only way we could improve it would require wrapping the longs into a separate parameter class , but then the constructor of this parameter class would have the same ugliness 😜  And it's the responsibility of the user of this class to correctly call this constructor in a sub class and to write a test for it.

lauraluiz:
@katmatt are then the new changes good? Shall we merge? @acbeni gave green light on his side.

lauraluiz:
(One build failed because of the search error in OSGi as expected, the other one is re-running for 13234534th time, hoping this time it won't get stuck as per usual 😓

katmatt:
@lauraluiz I'm fine with the changes and you can merge it 👍

PR #1465, opened 2017-07-05
Initial commit

### LINKED ISSUE #1462: Add new  removedImageUrls property to ProductPublishedMessage
Our API added the `removedImageUrls` property to the ProductPublished message [ProductPublished Message](http://dev.commercetools.com/http-api-projects-messages.html#productpublished-message) and we have to add the property to our `ProductPublishedMessage` class.

acbeni:
closes #1462

### DIFF
--- a/commercetools-models/src/main/java/io/sphere/sdk/products/messages/ProductPublishedMessage.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/products/messages/ProductPublishedMessage.java
@@ -18,13 +18,21 @@ public final class ProductPublishedMessage extends GenericMessageImpl<Product> {
 
     private final ProductProjection productProjection;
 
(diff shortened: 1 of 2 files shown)


commercetools-models/src/main/java/io/sphere/sdk/products/messages/ProductPublishedMessage.java
```diff
@@ -18,13 +18,21 @@
 
     private final ProductProjection productProjection;
 
+    private final String[] removedImageUrls;
```
katmatt: We should use `List<String>` here instead of an array.

acbeni: Done 👍

PR #1471, opened 2017-07-17
Revapi check goal, close #1432

### LINKED ISSUE #1432: Add revapi maven check goal to build
Currently we are using the revapi maven goal `report` when we release our SDK. The revapi maven plugin provides the goal `check`, which fails the build if there was a breaking change to our public API that isn't reflected in the current version.

We should try to integrate the `check` goal into our build, so that we have to update the project version before we do a pull request. This would help us in releasing our SDK, because the project version on master would then conform to the semantic versioning scheme enforced by the `check` goal.

References:

- [Revapi Project](http://revapi.org/)
- [Revapi Maven Plugin](http://revapi.org/modules/revapi-maven-plugin/index.html)
- [Revapi Maven Plugin `check` goal](http://revapi.org/modules/revapi-maven-plugin/check-mojo.html)

ccsalazarr-commercetools:
@katmatt, now it is generated a report per module and one aggregated report during the package phase. But I didn't execute the check goal, because the build will fail (when there are breaking changes) no matter if the version is updated.

Could you review it please?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,8 +64,8 @@
     <scm>
         <url>git@github.com:commercetools/commercetools-jvm-sdk.git</url>
         <connection>scm:git:git@github.com:commercetools/commercetools-jvm-sdk.git</connection>
-      <tag>HEAD</tag>
-  </scm>
+        <tag>HEAD</tag>
+    </scm>
     <properties>
         <jackson.version>2.6.5</jackson.version>
         <gson.version>2.5</gson.version>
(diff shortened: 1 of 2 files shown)


ccsalazarr-commercetools:
@katmatt, when I added the `report-aggregate` goal into the root POM.xml, the build takes a really long time and it fails in travis because of that, but when I use only `report` goal (to generate the report for each module), the build time is normal and it passes. I am trying to figure out a solution, but maybe do you have some advice?

katmatt:
@ccsalazarr-commercetools We could try to use a maven build profile to activate the `report-aggregate` goal. Let's discuss this further tomorrow.

pom.xml
```diff
@@ -363,6 +370,41 @@
...
+                    <newArtifacts>
+                        <artifact>${project.groupId}:${project.artifactId}:${project.version}</artifact>
+                    </newArtifacts>
+                </configuration>
+                <executions>
+                    <execution>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>report</goal>
+                            <goal>report-aggregate</goal>
```
katmatt: This line is problematic: the `report-aggregate` goal triggers the `package` phase and we end up in an infinite loop.

ccsalazarr-commercetools: I tried to add the `report-aggregate` within `install` or `site` phases, but it didn't work either. For now, I removed the problematic goal.

Issue #1491, opened 2017-08-01
Add correlation id generator


katmatt:
Our client should add a generate correlation id to every request. The correlation id should be set as `commercetools-jvm-sdk` http header. The format of the correlation id is defined as:

```
correlationId := <projectKey> '/' <UUID> ('/' <FURTHER_ID)*
```

PR #1492, opened 2017-08-01
Add correlation id generator

### LINKED ISSUE #1131: evaluate generating the correlation ID on the client side

katmatt:
Closes #1131 #1491

### DIFF
--- a/commercetools-java-client-core/src/test/java/io/sphere/sdk/client/SphereClientTest.java
+++ b/commercetools-java-client-core/src/test/java/io/sphere/sdk/client/SphereClientTest.java
@@ -55,14 +55,12 @@ public class SphereClientTest {
             final DummySphereRequest sphereRequest = DummySphereRequest.of();
 
             final CompletableFuture<String> completionStage = (CompletableFuture<String>) sphereClient.execute(sphereRequest);
-            final String response = completionStage.get();
-
-            assertThat(response).isEqualTo(DummySphereRequest.DEFAULT_RESPONSE_OBJECT);
+            assertThat(completionStage).isCompletedWithValue(DummySphereRequest.DEFAULT_RESPONSE_OBJECT);
 
             final HttpHeaders headers = httpRequestArgumentCaptor.getValue().getHeaders();
             final Optional<String> correlationId = headers.findFlatHeader(HttpHeaders.X_CORRELATION_ID);
 
-            assertThat(correlationId.isPresent()).isTrue();
+            assertThat(correlationId).isPresent();
 
             final List<String> correlationIdParts = Arrays.asList(correlationId.get().split("/"));
 
--- a/commercetools-java-client-core/src/test/java/io/sphere/sdk/client/SphereClientTest.java
+++ b/commercetools-java-client-core/src/test/java/io/sphere/sdk/client/SphereClientTest.java
@@ -43,7 +43,6 @@ public class SphereClientTest {
         when: {
             when(sphereApiConfig.getProjectKey()).thenReturn(projectKey);
             when(sphereApiConfig.getApiUrl()).thenReturn("http://api.commercetools.de");
-            when(sphereApiConfig.getCorrelationIdGenerator()).thenReturn(CorrelationIdGenerator.of(projectKey));
             when(httpClient.getUserAgent()).thenReturn("user-agent");
             when(sphereAccessTokenSupplier.get()).thenReturn(CompletableFutureUtils.successful("token"));
 
--- a/commercetools-java-client-core/src/test/java/io/sphere/sdk/client/SphereClientTest.java
+++ b/commercetools-java-client-core/src/test/java/io/sphere/sdk/client/SphereClientTest.java
@@ -55,7 +55,8 @@ public class SphereClientTest {
             final DummySphereRequest sphereRequest = DummySphereRequest.of();
 
             final CompletableFuture<String> completionStage = (CompletableFuture<String>) sphereClient.execute(sphereRequest);
-            assertThat(completionStage).isCompletedWithValue(DummySphereRequest.DEFAULT_RESPONSE_OBJECT);
+            final String response = completionStage.get();
+            assertThat(response).isEqualTo(DummySphereRequest.DEFAULT_RESPONSE_OBJECT);
 
             final HttpHeaders headers = httpRequestArgumentCaptor.getValue().getHeaders();
             final Optional<String> correlationId = headers.findFlatHeader(HttpHeaders.X_CORRELATION_ID);
--- a/commercetools-java-client-core/src/test/java/io/sphere/sdk/client/SphereClientTest.java
+++ b/commercetools-java-client-core/src/test/java/io/sphere/sdk/client/SphereClientTest.java
@@ -1,0 +1,78 @@
+package io.sphere.sdk.client;
+
+import io.sphere.sdk.http.HttpClient;
+import io.sphere.sdk.http.HttpHeaders;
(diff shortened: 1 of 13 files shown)


katmatt:
@schleichardt Looking forward to your feedback 😄

commercetools-sdk-base/src/main/java/io/sphere/sdk/client/CorrelationIdGenerator.java
```diff
@@ -0,0 +1,31 @@
+package io.sphere.sdk.client;
+
+/**
+ * Generates correlation ids for each request.
+ *
+ * The format of the correlation id is defined as:
+ * {@code <projectKey> / <UUID> (/ customId)*}
+ */
+public interface CorrelationIdGenerator {
```
schleichardt: concerning the documentation it has too much implementation detail. I would just expect from the interface to say it will generate a unique correlationId and has only the createCorrelationId method.

commercetools-sdk-base/src/main/java/io/sphere/sdk/client/CorrelationIdGenerator.java
```diff
@@ -0,0 +1,31 @@
+package io.sphere.sdk.client;
+
+/**
+ * Generates correlation ids for each request.
+ *
+ * The format of the correlation id is defined as:
+ * {@code <projectKey> / <UUID> (/ customId)*}
```
schleichardt: I think this docu belongs to the implementation.

commercetools-sdk-base/src/main/java/io/sphere/sdk/client/SphereApiConfig.java
```diff
@@ -10,11 +10,17 @@
 
     String getProjectKey();
 
+    CorrelationIdGenerator getCorrelationIdGenerator();
```
schleichardt: I would not expose this method in SphereApiConfig, you would break a lot of my projects by additionally requiring to implement this method which the user may not care about.
If you disagree to expose this, at least implement it as default method so you don't break the clients.

commercetools-sdk-base/src/main/java/io/sphere/sdk/client/CorrelationIdGenerator.java
schleichardt: Also it is not necessary to expose this interface, I don't see it as use case that clients want to override this. Even if they want to override it, they could use a SphereClientDecorator.

commercetools-sdk-base/src/test/java/io/sphere/sdk/client/CorrelationIdGeneratorTest.java
```diff
@@ -0,0 +1,33 @@
...
+    public void createCorrelationId() {
+        final String projectKey = "my-project";
+
+        final CorrelationIdGenerator correlationIdGenerator =
+                CorrelationIdGenerator.of(projectKey);
+
+        assertThat(correlationIdGenerator.getProjectKey()).isEqualTo(projectKey);
+
+        final String correlationId = correlationIdGenerator.createCorrelationId();
+        final List<String> correlationIdParts = Arrays.asList(correlationId.split("/"));
```
schleichardt: I think more readable would be something like 
```
assertThat(correlationId).matches("my-project/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}")
```

katmatt: That's a good idea 👍

commercetools-sdk-base/src/main/java/io/sphere/sdk/client/SphereApiConfig.java
katmatt: Good point 👍  I like your suggestion and will incorporate it into my changes!

commercetools-sdk-base/src/main/java/io/sphere/sdk/client/CorrelationIdGenerator.java
katmatt: I agree that users could write their own decorator, but still think it would be useful to expose this interface so that our users can customize the correlation id generator. And I think exposing this interface makes it easier to customize the correlation id generation. I'll think about it further and then we can discuss it.

commercetools-java-client-core/src/test/java/io/sphere/sdk/client/SphereClientTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+            final CompletableFuture<String> completionStage = (CompletableFuture<String>) sphereClient.execute(sphereRequest);
+            final String response = completionStage.get();
+
+            assertThat(response).isEqualTo(DummySphereRequest.DEFAULT_RESPONSE_OBJECT);
+
+            final HttpHeaders headers = httpRequestArgumentCaptor.getValue().getHeaders();
+            final Optional<String> correlationId = headers.findFlatHeader(HttpHeaders.X_CORRELATION_ID);
+
+            assertThat(correlationId.isPresent()).isTrue();
```
schleichardt: an alternative:

```
assertThat(correlationId).isPresent();
```

commercetools-java-client-core/src/test/java/io/sphere/sdk/client/SphereClientTest.java
```diff
@@ -0,0 +1,77 @@
...
+        }
+        then:
+        {
+            final SphereClient sphereClient = SphereClient.of(sphereApiConfig, httpClient, sphereAccessTokenSupplier);
+            final DummySphereRequest sphereRequest = DummySphereRequest.of();
+
+            final CompletableFuture<String> completionStage = (CompletableFuture<String>) sphereClient.execute(sphereRequest);
+            final String response = completionStage.get();
+
+            assertThat(response).isEqualTo(DummySphereRequest.DEFAULT_RESPONSE_OBJECT);
```
schleichardt: briefer:

```
assertThat(completionStage).isCompletedWithValue(DummySphereRequest.DEFAULT_RESPONSE_OBJECT);
```

commercetools-sdk-base/src/main/java/io/sphere/sdk/client/SphereClientConfig.java
```diff
@@ -78,6 +80,11 @@ public String getProjectKey() {
         return projectKey;
     }
 
+
+    public CorrelationIdGenerator getCorrelationIdGenerator() {
```
schleichardt: Why do you need here a getter?

katmatt: This is needed by the `SphereClientConfigBuilder`, but I reduced the visibility to default.

commercetools-java-client-core/src/test/java/io/sphere/sdk/client/SphereClientTest.java
katmatt: Good idea 👍

commercetools-java-client-core/src/test/java/io/sphere/sdk/client/SphereClientTest.java
katmatt: Looks like I didn't know about this, thx for sahring it with me =)

lauraluiz: New to me too, and a very much needed assertion! :D

PR #1503, opened 2017-08-10
Key property customer, close #1496

### LINKED ISSUE #1496: Add key property to Customer
Our `Customer` and `CustomerDraft` types should support the new `key`property and both of them should extend our `WithKey` interface. Additional we should add the corresponding new `SetKey` update action. Deleting and Updating by `key`should be added too.

ccsalazarr-commercetools:


commercetools-models/src/main/java/io/sphere/sdk/customers/Customer.java
```diff
@@ -120,28 +120,42 @@
...
      * @see io.sphere.sdk.carts.queries.CartByCustomerIdGet
+     *
      * @return ID
      */
     @HasNoUpdateAction
     @Override
     String getId();
 
     /**
+     * User-specific unique identifier for a customer. Must be unique across a project.
```
katmatt: Nice javadoc 👍

PR #1510, opened 2017-08-25
Add shipping method get and update by key.

### LINKED ISSUE #1505: Add get and update by key for ShippingMethod
We forgot to add support for getting and updating a `ShippingMethod` by key when we worked on #1475.

katmatt:
Closes #1505


katmatt:
@lauraluiz That's the adavantage of our code generators 😄  And when we start to generate our test fixture we may further reduce the number of lines in the future 😉

Issue #1512, opened 2017-08-29
QueryExecutionUtils.queryAll performance improvement


heshamMassoud:
Currently the JVM SDK provides a convenience implementation for fetching all elements of a resource.

````java
QueryExecutionUtils.queryAll(sphereClient, query)
````
However, it keeps all the results for every page in memory. 

This, however, could be improved (performance-wise) in the sense that if a user wants to only fetch all product keys, the queryAll would need instead to store all product data in memory and the user would then extract all the keys from them. This would causes unnecessary memory usage from the app using JVM SDK.
 
Therefore, an alternative implementation would be to have a callback and/or supplier that is called on every page to do the processing for this page, then remove it from memory.

heshamMassoud:
Currently preparing a PR for the proposed solution.

heshamMassoud:
PR openned here: #1513

andrii-kovalenko-ct:
My suggestion is to make consumer/function callback not on every page, but on every item, and incapsulate the iteration in the query all utility method.

Literally i mean:

  1.  Instead of
 `... run(... final Function<List<T>, S> callback) {` 
    should be
`... run(... final Function<T, S> callback) {` 

```java
<S> CompletionStage<List<S>> run(final SphereClient client, final Function<T, S> resultMapper) {
    return queryPage(client, 0).thenCompose(result -> {
        final List<CompletableFuture<List<S>>> futureResults = new ArrayList<>();

        List<S> callbackResult = result.getResults().stream().map(resultMapper).collect(toList());

        List<CompletableFuture<List<S>>> nextPagesCallbackResults = queryNextPages(client, result.getTotal(), resultMapper);

        futureResults.add(completedFuture(callbackResult));
        futureResults.addAll(nextPagesCallbackResults);

        return transformListOfFuturesToFutureOfLists(futureResults);
    });
}

private <S> List<CompletableFuture<List<S>>> queryNextPages(final SphereClient client, final long totalElements,
                                                            final Function<T, S> resultMapper) {
    final long totalPages = getTotalNumberOfPages(totalElements);
    return LongStream.range(1, totalPages)
            .mapToObj(page -> queryPage(client, page)
                    .thenApply(PagedQueryResult::getResults)
                    .thenApply(results -> results.stream().map(resultMapper).collect(toList()))
                    .toCompletableFuture())
            .collect(toList());
}

private <S> CompletableFuture<List<S>> transformListOfFuturesToFutureOfLists(
[...]

PR #1513, opened 2017-08-30
Query all improvement

heshamMassoud:
Addresses issue: #1512 by adding callbacks as either:
1.  `Function` for returning a result from every page (for example list of external ids for every page of categories, as seen [here](https://github.com/commercetools/commercetools-jvm-sdk/pull/1513/files#diff-041894d3b01cc131f14b780b4d036ecdR116) ) 
 or
2.  `Consumer` for not returning results from every page, but `void`. This would be normally used for only processing pages but not returning. E.g. collecting external ids in a preexisting list (as seen [here](https://github.com/commercetools/commercetools-jvm-sdk/pull/1513/files#diff-041894d3b01cc131f14b780b4d036ecdR138))

Also a bug that was noticed in the old `QueryAllImpl` was in the `queryNextPages`  during development was that there was always an extra empty page fetched. So for example: if there are 4 categories and you fetch all with pageSize 2. It means:
````java
final long totalPages = totalElements / pageSize; // 4/2 = 2
return LongStream.rangeClosed(1, totalPages) // range is start and end inclusive (so this loops on 1,2 incl.)
````

and since page 0 was already fetched in advance before the `queryNextPages` was called. So It means in total you would fetch 0,1,2 (3 pages), when only 2 pages where needed.

This PR, however, fixes that with the `getTotalNumberofPages()` method and using `LongStream.range` which is start inclusive, but end exclusive. [here](https://github.com/commercetools/commercetools-jvm-sdk/pull/1513/files#diff-763fe77a9180a332bb50b896cff258d7R105)

Integration and unit tests were also added.

### CODE REVIEW

**Comment by katmatt** [comment]
😎  Looking forward to review your PR soon

File: commercetools-convenience/src/main/java/io/sphere/sdk/queries/QueryAllImpl.java [hunk 1]
```diff
@@ -22,42 +24,140 @@ private QueryAllImpl(final QueryDsl<T, C> baseQuery, final long pageSize) {
         this.pageSize = pageSize;
     }
 
-    public CompletionStage<List<T>> run(final SphereClient client) {
+    /**
+     * Given a {@code callback} {@link Function}, this method applies this callback on each page of the results from
```
katmatt: Although this class isn't visible to our users, I like that you added javadoc here 👍

commercetools-convenience/src/main/java/io/sphere/sdk/queries/QueryExecutionUtils.java
```diff
@@ -2,11 +2,15 @@
...
 import java.util.concurrent.CompletionStage;
+import java.util.function.Consumer;
+import java.util.function.Function;
 
 /**
  * Provides facilities to fetch all elements matching a query predicate.
+ *
  * @since 1.0.0-RC1
  */
 public final class QueryExecutionUtils {
```
katmatt: Since this class is public, we have to keep the previous method signatures. So we should add new methods for your extension.

commercetools-convenience/src/test/java/io/sphere/sdk/queries/QueryExecutionUtilsTest.java
```diff
@@ -51,9 +51,24 @@ public void onUnsortedResponses() throws Exception {
         withClient(clientWithDelays(16), list -> assertThat(list).isSortedAccordingTo(categoryComparator).hasSize(16));
     }
 
+    @Test
+    public void onGetTotalNumberOfPages() {
```
katmatt: Great that you added a test for this method 👍

commercetools-convenience/src/test/java/io/sphere/sdk/meta/CategoryDocumentationIntegrationTest.java
```diff
@@ -299,15 +381,25 @@ private static void withProductInCategory(final BlockingSphereClient client, fin
...
+            .blockingWait(categoryPagesStage, Duration.ofMinutes(5));
+        final List<Category> categories = categoryPages.stream()
+                                                       .flatMap(List::stream)
+                                                       .collect(toList());
         return CategoryTree.of(categories);
     }
 
     private CategoryTree createCategoryTree() {
         //stuff from previous example
-        final CompletionStage<List<Category>> categoriesStage = QueryExecutionUtils.queryAll(client(), CategoryQuery.of());
```
katmatt: When you changed the code to also provide the old API, you should be able to revert these changes.

commercetools-convenience/src/test/java/io/sphere/sdk/meta/CategoryDocumentationIntegrationTest.java
```diff
@@ -299,15 +381,25 @@ private static void withProductInCategory(final BlockingSphereClient client, fin
     }
 
     private static CategoryTree fetchCurrentTree() {
-        final CompletionStage<List<Category>> categoriesStage = QueryExecutionUtils.queryAll(client(), CategoryQuery.of());
```
katmatt: When you changed the code to also provide the old API, you should be able to revert these changes.

commercetools-convenience/src/test/java/io/sphere/sdk/meta/CategoryDocumentationIntegrationTest.java
```diff
@@ -73,29 +74,110 @@ public static void beforeClass() throws IOException {
...
-                .matches(cats -> cats.parallelStream().anyMatch(cat -> cat.get("slug").get("en").asText().equals("boots-women")));
+            .hasSize(15)
+            .matches(cats -> cats.parallelStream()
+                                 .anyMatch(cat -> cat.get("slug").get("en").asText().equals("boots-women")));
     }
 
     @Test
     public void fetchRoots() throws Exception {
         final CategoryQuery seedQuery = CategoryQuery.of().withPredicates(m -> m.parent().isNotPresent());
-        final CompletionStage<List<Category>> categoriesStage = QueryExecutionUtils.queryAll(client(), seedQuery);
```
katmatt: When you changed the code to also provide the old API, you should be able to revert these changes.

commercetools-convenience/src/test/java/io/sphere/sdk/meta/CategoryDocumentationIntegrationTest.java
```diff
@@ -73,29 +74,110 @@ public static void beforeClass() throws IOException {
...
+
         assertThat(categories)
-                .hasSize(15)
-                .matches(cats -> cats.parallelStream().anyMatch(cat -> cat.getSlug().get(ENGLISH).equals("boots-women")));
+            .hasSize(15)
+            .matches(cats -> cats.parallelStream().anyMatch(cat -> cat.getSlug().get(ENGLISH).equals("boots-women")));
     }
+
+    @Test
+    public void fetchAllExternalIdsWithUniformPageSizes() throws Exception {
```
katmatt: Nice that you added tests that show/document your use case 👍

commercetools-convenience/src/main/java/io/sphere/sdk/queries/QueryAllImpl.java
```diff
@@ -22,42 +24,140 @@ private QueryAllImpl(final QueryDsl<T, C> baseQuery, final long pageSize) {
...
+    @Nonnull
+    private <S> CompletableFuture<List<S>> transformListOfFuturesToFutureOfLists(
+        final List<CompletableFuture<S>> futures) {
         final CompletableFuture[] futuresAsArray = futures.toArray(new CompletableFuture[futures.size()]);
         return CompletableFuture.allOf(futuresAsArray)
-                .thenApply(x -> futures.stream()
-                        .map(CompletableFuture::join)
-                        .flatMap(Collection::stream)
-                        .collect(Collectors.<T>toList()));
+                                .thenApply(x -> futures.stream()
```
katmatt: I don't fully understand this line. Why do we first convert the futures to an array and then then later use list of futures `x -> futures.stream()` ? What am I missing here? 

But looks like this code was there before you extended it 😉  But would be nice to understand it anyway 😄

commercetools-convenience/src/main/java/io/sphere/sdk/queries/QueryAllImpl.java
```diff
@@ -22,42 +24,140 @@ private QueryAllImpl(final QueryDsl<T, C> baseQuery, final long pageSize) {
...
+
+    /**
+     * Given a total number of elements {@code totalElements} this method calculates the number of pages needed that
+     * would be needed to cover all the elements with set {@code pageSize} of {@link this} instance.
+     *
+     * @param totalElements number of elements to get the number of pages for.
+     * @return the total number of pages.
+     */
+    long getTotalNumberOfPages(final long totalElements) {
+        return (long) Math.ceil((double) totalElements / pageSize);
```
katmatt: In Java the `/` operator has a higher predence that the cast operator. That means that this code performs an integer division and afterwards does the cast.
The correct code looks like: `Math.ceil(((double) totalElements) / pageSize)`

Would be nice if you could add some more test cases for this method 😉

andrii-kovalenko-ct: > In Java the / operator has a higher predence that the cast operator. 

actually, no, [cast has higher priority](http://introcs.cs.princeton.edu/java/11precedence/)

andrii-kovalenko-ct: [But don't re-invent the wheel](https://www.google.de/search?client=opera&q=java+pagination+calculate+number+pages&oq=java+pagination+calculate+number+pages&gs_l=psy-ab.3...274434.274841.0.275577.5.5.0.0.0.0.244.376.0j1j1.2.0....0...1.1.64.psy-ab..4.0.0._N9H518yYzw):

```java
int pages = (totalElements + pageSize - 1) / pageSize;
```

commercetools-convenience/src/main/java/io/sphere/sdk/queries/QueryAllImpl.java
andrii-kovalenko-ct: because of JDK function signature: `java.util.concurrent.CompletableFuture#allOf(CompletableFuture<?>... cfs)`  - it accepts an array, not a list. But later in `thenApply` it is nicer to work with List/Stream abstraction, instead of raw arrays.

Not sure if it is possible to simplify it here.

commercetools-convenience/src/test/java/io/sphere/sdk/queries/QueryExecutionUtilsTest.java
andrii-kovalenko-ct: corner cases are missing, like:
  - [ ] `totalElements = 0`
  - [ ] `totalElements == pageSize`
    - [ ] `totalElements == pageSize - 1`
    - [ ] `totalElements == pageSize + 1`
  - [ ] `pageSize == 1`

commercetools-convenience/src/test/java/io/sphere/sdk/meta/CategoryDocumentationIntegrationTest.java
```diff
@@ -73,29 +74,110 @@ public static void beforeClass() throws IOException {
...
+            .blockingWait(categoryPagesStage, Duration.ofMinutes(5));
+
+        assertThat(categoryPages).hasSize(4);
+
+        final List<String> externalIds = categoryPages.stream()
+                                                      .flatMap(List::stream)
+                                                      .collect(toList());
+
+        assertThat(externalIds).hasSize(15);
+        IntStream.range(0, externalIds.size()).forEach(index -> assertThat(externalIds).contains(index + ""));
```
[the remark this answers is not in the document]
katmatt: @andrii-kovalenko-ct That's a good idea and I'm the right person to ask 😉 And since it's only used in our tests the update won't cause any problems for our users. I've created a more general issue for updating our test dependencies and you can find it here: #1518

commercetools-convenience/src/main/java/io/sphere/sdk/queries/QueryAllImpl.java
katmatt: @andrii-kovalenko-ct I see, it's because `CompletableFuture.allOf` returns `CompletionStage<Void>`

katmatt:
@heshamMassoud Any update on this? We always try to limit the number of open pull request and thus it would be nice to get this PR closed 😉

heshamMassoud:
Hi @katmatt, thanks a lot for review (and @andrii-kovalenko-ct of course) I want to respond to the review. But I am really busy with another topic until the end of next week for a release. So after It I will have time for it.

commercetools-convenience/src/main/java/io/sphere/sdk/queries/QueryAllImpl.java
butenkor: @andrii-kovalenko-ct Not sure how is it better? You will have to round in both implementations.

andrii-kovalenko-ct: the question not about rounding: this is just less confusing, has less casting and no `Math` method calls.

But talking precisely: in the last proposition there is no rounding, it is completely integer arithmetics.

andrii-kovalenko-ct:
> Any update on this? We always try to limit the number of open pull requests and thus it would be nice to get this PR closed 

@katmatt i'm taking care about this PR now.

katmatt:
@andrii-kovalenko-ct Ok 👍

commercetools-convenience/src/main/java/io/sphere/sdk/queries/QueryAllImpl.java
andrii-kovalenko-ct: 463ce43f13

commercetools-convenience/src/main/java/io/sphere/sdk/queries/QueryExecutionUtils.java
andrii-kovalenko-ct: reverted legacy API: 16d5ea6

commercetools-convenience/src/test/java/io/sphere/sdk/queries/QueryExecutionUtilsTest.java
andrii-kovalenko-ct: 63b893d

commercetools-convenience/src/test/java/io/sphere/sdk/meta/CategoryDocumentationIntegrationTest.java
andrii-kovalenko-ct: reverted legacy API: 16d5ea6

commercetools-convenience/src/test/java/io/sphere/sdk/meta/CategoryDocumentationIntegrationTest.java
andrii-kovalenko-ct: reverted legacy API: 16d5ea6

commercetools-convenience/src/test/java/io/sphere/sdk/meta/CategoryDocumentationIntegrationTest.java
andrii-kovalenko-ct: reverted legacy API: 16d5ea6

commercetools-convenience/src/test/java/io/sphere/sdk/meta/CategoryDocumentationIntegrationTest.java
```diff
@@ -73,29 +74,110 @@ public static void beforeClass() throws IOException {
...
+    public void fetchAllExternalIdsWithUniformPageSizes() throws Exception {
+        final Function<List<Category>, List<String>> externalIdCallBack = categories ->
+            categories.stream()
+                      .map(Category::getExternalId)
+                      .collect(toList());
+
+        final CompletionStage<List<List<String>>> categoryPagesStage = QueryExecutionUtils
+            .queryAll(client(), CategoryQuery.of(), externalIdCallBack, 3);
+        final List<List<String>> categoryPages = SphereClientUtils
+            .blockingWait(categoryPagesStage, Duration.ofMinutes(5));
```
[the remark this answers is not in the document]
andrii-kovalenko-ct: 6b0c32c239

andrii-kovalenko-ct:
@katmatt @heshamMassoud @butenkor @acbeni 
i've re-factored the PR, the main changes:
  - reverted previous legacy functions signatures (e.g. `QueryAllImpl#run(SphereClient)`, `QueryExecutionUtils#queryAll(SphereClient, QueryDsl<T,C>)`)
  - instead of mapping pages to lists of results, the pages iteration is encapsulated in `QueryAllImpl`, and the values mapper/consumer work with entities themselves, not lists
  - flatten most of the CompletionStages
  - flatten Streams (avoid Stream->toList->toStream converting chains)

I'd say the branch is completely re-factored so it is recommended to review it from scratch.

andrii-kovalenko-ct:
Also, there is one point which i still have doubts:
should we collect the pages results in parallel streams, or should be do this sequentially?

So far i stopped on sequential:
https://github.com/commercetools/commercetools-jvm-sdk/pull/1513/files#diff-763fe77a9180a332bb50b896cff258d7R51

https://github.com/commercetools/commercetools-jvm-sdk/pull/1513/files#diff-763fe77a9180a332bb50b896cff258d7R73

The reason is:
since JDK does not provide any way to make streams parallelization in dedicated thread pools (default `ForkJoinPool` is used always) and if we have a lot of pages to request, it might be that we completely block the `ForkJoinPool` by our paging fetching, and all other processes inside same ClassLoader (e.g. in the same application/web/docker-container) will be blocked as well.

See more about the topic:
https://dzone.com/articles/think-twice-using-java-8

WDYT: **should we try to parallelize the process**? Or we could stick with sequential solution so far and extend it in the future with custom thread pool executors?

commercetools-convenience/src/main/java/io/sphere/sdk/queries/QueryAllImpl.java
```diff
@@ -22,42 +22,125 @@ private QueryAllImpl(final QueryDsl<T, C> baseQuery, final long pageSize) {
...
+     * of this {@code resultsMapper} on each item.
+     *
+     * @param client        the CTP client that the query is run on.
+     * @param resultsMapper the callback that gets called on each item of results.
+     * @param <S>           the type of the result mapped values of the callback on each requested {@code T} item.
+     * @return a future containing a list of results of the {@code resultsMapper} on each item.
+     */
+    @Nonnull
+    <S> CompletionStage<List<S>> run(final SphereClient client, final Function<T, S> resultsMapper) {
+        return queryPage(client, 0).thenApply(result -> {
```
katmatt: I personally would extract this lambda into a separate method and then use a method handle instead of the lambda. But it's a matter of taste and we can keep your code as it is 😉

commercetools-convenience/src/main/java/io/sphere/sdk/queries/QueryExecutionUtils.java
```diff
@@ -29,14 +34,91 @@ private QueryExecutionUtils() {
...
+     * @param <T>      type of one query result element
+     * @param <C>      type of the query
      * @return elements
      */
     public static <T, C extends QueryDsl<T, C>> CompletionStage<List<T>> queryAll(final SphereClient client, final QueryDsl<T, C> query, final int pageSize) {
         return QueryAllImpl.of(query, pageSize).run(client);
     }
+
+    /**
+     * Queries all elements matching a query by using an offset based pagination with page size 500.
```
katmatt: Nice javadoc 👍

commercetools-convenience/src/test/java/io/sphere/sdk/meta/CategoryDocumentationIntegrationTest.java
```diff
@@ -74,31 +74,105 @@ public static void beforeClass() throws IOException {
...
     public void fetchAll() throws Exception {
         final CompletionStage<List<Category>> categoriesStage = QueryExecutionUtils.queryAll(client(), CategoryQuery.of(), 500);
-        final List<Category> categories = SphereClientUtils.blockingWait(categoriesStage, Duration.ofMinutes(5));
+        final List<Category> categories = SphereClientUtils.blockingWait(categoriesStage, Duration.ofMinutes(1));
         assertThat(categories)
                 .hasSize(15)
                 .matches(cats -> cats.parallelStream().anyMatch(cat -> cat.getSlug().get(ENGLISH).equals("boots-women")));
     }
+
+    @Test
```
katmatt: I like your tests 👍

andrii-kovalenko-ct: the tests are mostly done by @heshamMassoud though :)

I just re-factored the tests after re-factored the solution :)

commercetools-convenience/src/main/java/io/sphere/sdk/queries/QueryAllImpl.java
andrii-kovalenko-ct: completely agree: db35382

andrii-kovalenko-ct:
@katmatt 
> I would suggest to stick to the sequential flow

This is also my opinion!

> Another interesting option is to think about integrating the reactive streams from Java 9

OMG, i reviewed a tone of articles about Java9 release, and missed this feature. Anyway - this is for future :)

PR #1528, opened 2017-09-13
Add custom fields to DiscountCode.

### LINKED ISSUE #1526: Add custom fields to DiscountCode
Our `DiscountCode` type now support custom fields and we have to add support for it.

katmatt:
Closes #1526

### DIFF
--- a/commercetools-models/src/main/java/io/sphere/sdk/discountcodes/commands/updateactions/SetCustomField.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/discountcodes/commands/updateactions/SetCustomField.java
@@ -1,0 +1,34 @@
+package io.sphere.sdk.discountcodes.commands.updateactions;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import io.sphere.sdk.discountcodes.DiscountCode;
+import io.sphere.sdk.json.SphereJsonUtils;
+import io.sphere.sdk.types.customupdateactions.SetCustomFieldBase;
+
+/**
+ * Updates a custom field.
+ *
+ * {@doc.gen intro}
+ *
+ * {@include.example io.sphere.sdk.discountcodes.DiscountCodeCustomFieldsIntegrationTest#createDiscountCodeWithCustomType()}
+ *
+ * @see io.sphere.sdk.types.Custom
+ */
+public final class SetCustomField extends SetCustomFieldBase<DiscountCode> {
+    private SetCustomField(final String name, final JsonNode value) {
+        super(name, value);
+    }
+
+    public static SetCustomField ofJson(final String name, final JsonNode value) {
+        return new SetCustomField(name, value);
+    }
+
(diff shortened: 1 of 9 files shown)


katmatt:
@acbeni I improved the new tests and the failing tests are our usual flaky tests. That's why I would suggest to do the code review although our tests are failing at the moment.

commercetools-models/src/main/java/io/sphere/sdk/discountcodes/commands/updateactions/SetCustomField.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Updates a custom field.
+ *
+ * {@doc.gen intro}
+ *
+ * {@include.example io.sphere.sdk.discountcodes.DiscountCodeCustomFieldsIntegrationTest#createDiscountCodeWithCustomType()}
+ *
+ * @see io.sphere.sdk.types.Custom
+ */
+public final class SetCustomField extends SetCustomFieldBase<DiscountCode> {
```
acbeni: in most classes that extends `SetCustomFieldBase` we add methods `ofJson`, `ofObject` and `ofUnset`. wouldn't it make sense to add it to `SetCustomFieldBase` 🤔  ?

commercetools-models/src/main/java/io/sphere/sdk/discountcodes/commands/updateactions/SetCustomType.java
```diff
@@ -0,0 +1,45 @@
...
+/**
+ *
+ * Sets/unsets the custom type.
+ *
+ * {@doc.gen intro}
+ *
+ * {@include.example io.sphere.sdk.discountcodes.DiscountCodeCustomFieldsIntegrationTest#setCustomType()}
+ *
+ */
+public final class SetCustomType extends SetCustomTypeBase<DiscountCode> {
```
acbeni: Same for `SetCustomTypeBase`, wouldn't it be more convenient to add `ofTypeKeyAndObjects `, `ofTypeIdAndObjects`, `ofTypeIdAndJson `, `ofTypeKeyAndJson ` and `ofRemoveType` to `SetCustomTypeBase` ?

commercetools-models/src/main/java/io/sphere/sdk/discountcodes/commands/updateactions/SetCustomField.java
katmatt: They're static and always use the specific update action. And since our update action are specific to the resource type they're updating I currently don't see a better design 😞 Only solution that I see is a code generator for this very specific case. But that sounds like overkill to me. What do you think?

acbeni: Yeah the code generator solution seems like an overkill and after looking up there doesen't seem simple workaround 😢 
I would agree to let it that way 👍

PR #1536, opened 2017-09-15
Adding modlarity to sdk

acbeni:
This PR contains only the modularity part of the OSGI PR, the aim of this separation is to provide a kind of modularity that would be necessary for OSGI integration and as a preparation of the java 9 near release.


pom.xml
```diff
@@ -67,7 +67,7 @@
         <tag>HEAD</tag>
     </scm>
     <properties>
-        <jackson.version>2.6.5</jackson.version>
+        <jackson.version>2.8.9</jackson.version>
```
katmatt: This version update is a good idea, but according to our revapi check it introduces some incompatible changes. I'm fine with them, but I will explicitly mention this version update in our release notes.

pom.xml
```diff
@@ -102,6 +102,14 @@
         <maven-surefire-plugin.version>2.19.1</maven-surefire-plugin.version>
         <maven-deploy-plugin.version>2.8.2</maven-deploy-plugin.version>
         <handlebars.version>4.0.6</handlebars.version>
+        <pax-exam.version>4.11.0</pax-exam.version>
```
katmatt: Should we already merge this OSGi specific changes? What do you think?

pom.xml
```diff
@@ -102,6 +102,14 @@
...
+        <pax-exam.logging.version>1.10.1</pax-exam.logging.version>
+        <maven-jar-plugin.version>2.6</maven-jar-plugin.version>
+        <felix.maven.version>3.3.0</felix.maven.version>
```
katmatt: Should we already merge this OSGi specific changes? What do you think?

pom.xml
```diff
@@ -387,6 +435,29 @@
                 </executions>
             </plugin>
             <plugin>
+                <groupId>org.apache.felix</groupId>
```
katmatt: Should we already merge this OSGi specific changes? What do you think?

commercetools-internal-docs/src/main/java/io/sphere/sdk/meta/ReleaseNotes.java
```diff
@@ -207,7 +207,7 @@
  <li class=new-in-release>Added new messages {@link OrderPaymentStateChangedMessage}, {@link PaymentStatusInterfaceCodeSetMessage},
```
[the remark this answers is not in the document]
acbeni: Actually this file is identical to the current RelaseNotes file, i will see if i can remove it from this commit.

commercetools-java-client-ahc-1_8/src/main/java/io/sphere/sdk/client/SphereAsyncHttpClientFactory.java
```diff
@@ -4,13 +4,16 @@
 import io.sphere.sdk.http.AsyncHttpClientAdapter;
 import io.sphere.sdk.http.HttpClient;
 
-public final class SphereAsyncHttpClientFactory {
-    private SphereAsyncHttpClientFactory() {
-    }
+public final class SphereAsyncHttpClientFactory extends SphereHttpClientFactory{
 
     public static HttpClient create() {
```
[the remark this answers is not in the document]
acbeni: Yep that would make sense since it is candidate for removal in the next major release. 👍

commercetools-java-client-ahc-1_9/src/main/resources/META-INF/services/io.sphere.sdk.client.SphereHttpClientFactory
```diff
@@ -0,0 +1 @@
+io.sphere.sdk.client.SphereAsyncHttpClientFactory
```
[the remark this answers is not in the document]
acbeni: 👍  yep i've seen this issue, auomating more is always good 😄

commercetools-java-client-core/src/main/java/io/sphere/sdk/client/SphereHttpClientFactory.java
```diff
@@ -0,0 +1,11 @@
+package io.sphere.sdk.client;
+
+import io.sphere.sdk.http.HttpClient;
+import io.sphere.sdk.models.Base;
+
+
+public abstract class SphereHttpClientFactory extends Base{
```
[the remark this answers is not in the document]
acbeni: 👍  agreed and done.

pom.xml
acbeni: Yep i was also concerned about all the revapi complains, this is actually worth mentioning this in the release since it might create problems for people using jackson transitively from our API.

pom.xml
acbeni: 🤔  wasn't sur about it since it keeps track of only versions used in the osgi integration, but i after rethinking it it might be better to keep it separated in the osgi branch.

pom.xml
acbeni: Same as above
>🤔  wasn't sur about it since it keeps track of only versions used in the osgi integration, but i after rethinking it it might be better to keep it separated in the osgi branch.

pom.xml
acbeni: Same as above

>🤔 wasn't sur about it since it keeps track of only versions used in the osgi integration, but i after rethinking it it might be better to keep it separated in the osgi branch.

PR #1538, opened 2017-09-19
Fixing ci test runs

acbeni:
@katmatt  This is a fix for the SSL issue, can you please take a look at this PR.

### DIFF
--- a/commercetools-test-lib/src/main/java/io/sphere/sdk/test/ApacheHttpClientAdapterImpl.java
+++ b/commercetools-test-lib/src/main/java/io/sphere/sdk/test/ApacheHttpClientAdapterImpl.java
@@ -1,0 +1,133 @@
+package io.sphere.sdk.test;
+
+import io.sphere.sdk.http.*;
+import org.apache.commons.io.IOUtils;
+import org.apache.commons.io.input.AutoCloseInputStream;
+import org.apache.http.Header;
+import org.apache.http.HttpEntity;
+import org.apache.http.client.entity.UrlEncodedFormEntity;
+import org.apache.http.client.methods.HttpUriRequest;
+import org.apache.http.client.methods.RequestBuilder;
+import org.apache.http.entity.ContentType;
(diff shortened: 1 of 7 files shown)


commercetools-test-lib/src/main/java/io/sphere/sdk/test/ApacheHttpClientAdapterImpl.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.zip.GZIPInputStream;
+
+import static java.util.Arrays.asList;
+
+final class ApacheHttpClientAdapterImpl extends HttpClientAdapterBase {
```
katmatt: How about renaming this class to `IntegrationTestHttpClient`? And adding javadoc with a short explanation that we had to copy this class for our integration tests to avoid cyclic dependencies? What do you think?

commercetools-test-lib/src/main/java/io/sphere/sdk/test/CompletableFutureCallbackAdapter.java
```diff
@@ -0,0 +1,31 @@
+package io.sphere.sdk.test;
+
+
+
+import io.sphere.sdk.models.Base;
+import org.apache.http.concurrent.FutureCallback;
+
+import java.util.concurrent.CompletableFuture;
+
+final class CompletableFutureCallbackAdapter<T> extends Base implements FutureCallback<T> {
```
katmatt: I would suggest to add some javadoc that explains why we had to duplicate this class. What do you think?

acbeni:
@katmatt i agree with you, since by the time someone would try to delete the duplicated code without addressing the problem that led to it in the first place. 👍 
Both comments have been taking in code review

katmatt:
Looks good to me 👍

PR #1544, opened 2017-09-25
Fix NPE in priceDraftbuilder and add test for it.

### LINKED ISSUE #1543: NPE on PriceDraftBuilder of method
`customFields`  [is an optional field on a price](http://dev.commercetools.com/http-api-projects-products.html#price). Therefore, it could have the value `null`.

When building a `ProductVariantDraft` from a `ProductVariant` and one of the prices of the variant has null `customFields`, an NPE will be thrown by the method [`PriceDraftBuilder#of`](https://github.com/commercetools/commercetools-jvm-sdk/blob/master/commercetools-models/src/main/java/io/sphere/sdk/products/PriceDraftBuilder.java#L156) which will call the method [CustomFieldsDraftBuilder#of](https://github.com/commercetools/commercetools-jvm-sdk/blob/master/commercetools-models/src/main/java/io/sphere/sdk/types/CustomFieldsDraftBuilder.java#L57) causing the `customFields.getType()` call to throw the NPE.

katmatt:
Closes #1543

### DIFF
--- a/commercetools-models/src/main/java/io/sphere/sdk/products/PriceDraftBuilder.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/products/PriceDraftBuilder.java
@@ -147,12 +147,15 @@ public final class PriceDraftBuilder extends Base implements Builder<PriceDraftD
     }
 
     public static PriceDraftBuilder of(final Price template) {
+        final CustomFieldsDraft customFieldsDraft = Optional.ofNullable(template.getCustom())
+                .map(customFields -> CustomFieldsDraftBuilder.of(customFields).build())
+                .orElse(null);
         return of(template.getValue())
                 .country(template.getCountry())
                 .customerGroup(template.getCustomerGroup())
                 .channel(template.getChannel())
                 .validFrom(template.getValidFrom())
                 .validUntil(template.getValidUntil())
-                .custom(CustomFieldsDraftBuilder.of(template.getCustom()).build());
+                .custom(customFieldsDraft);
     }
 }
--- a/commercetools-models/src/main/java/io/sphere/sdk/products/PriceDraftBuilder.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/products/PriceDraftBuilder.java
@@ -146,16 +146,19 @@ public final class PriceDraftBuilder extends Base implements Builder<PriceDraftD
                 .custom(template.getCustom());
     }
 
-    public static PriceDraftBuilder of(final Price template) {
-        final CustomFieldsDraft customFieldsDraft = Optional.ofNullable(template.getCustom())
+    private static CustomFieldsDraft copyCustom(Price template) {
+        return Optional.ofNullable(template.getCustom())
                 .map(customFields -> CustomFieldsDraftBuilder.of(customFields).build())
                 .orElse(null);
+    }
+
+    public static PriceDraftBuilder of(final Price template) {
         return of(template.getValue())
                 .country(template.getCountry())
                 .customerGroup(template.getCustomerGroup())
                 .channel(template.getChannel())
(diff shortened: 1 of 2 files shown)


katmatt:
@heshamMassoud Would be great if you could do review the fix, so that we are sure that we really fixed your issue =)

commercetools-models/src/main/java/io/sphere/sdk/products/PriceDraftBuilder.java
```diff
@@ -147,12 +147,15 @@ public static PriceDraftBuilder of(final PriceDraft template) {
     }
 
     public static PriceDraftBuilder of(final Price template) {
+        final CustomFieldsDraft customFieldsDraft = Optional.ofNullable(template.getCustom())
```
heshamMassoud: @katmatt maybe the null-check logic (which is specific only for custom fields validation) should be in the method that's related to the customFieldsDraft only ->  [`custom()`](https://github.com/commercetools/commercetools-jvm-sdk/blob/93f960ce752b7951e81ed7bb7c3baa31c250be46/commercetools-models/src/main/java/io/sphere/sdk/products/PriceDraftBuilder.java#L75): https://github.com/commercetools/commercetools-jvm-sdk/blob/93f960ce752b7951e81ed7bb7c3baa31c250be46/commercetools-models/src/main/java/io/sphere/sdk/products/PriceDraftBuilder.java#L75

I see in this builder also Null checks are also made in their own specific methods, e.g. customerGroup, channel, etc..

katmatt: Let me see if I fully understand you:. You suggest to do it like this:
```

    public PriceDraftBuilder custom(@Nullable final CustomFields custom) {
        this.custom = Optional.ofNullable(template.getCustom())
                .map(customFields -> CustomFieldsDraftBuilder.of(customFields).build())
                .orElse(null);
        return this;
    }
```

I don't like this, because I don't want to expose an additional overridden `custom` method here. Especially since we just would introduce it for our copy method factories. And the other methods are more common conversions (from `Referenceable` to `Reference`) that are useful for more use cases. 

What do you think? I hope I didn't miss something in your comment?

heshamMassoud: I meant 

````java
 public PriceDraftBuilder custom(@Nullable final CustomFields custom) {
        this.custom = Optional.ofNullable(custom)
                .map(customFields -> CustomFieldsDraftBuilder.of(customFields).build())
                .orElse(null);
        return this;
    }
````
but why would you need to override the `custom` method here? Why not add this logic to the already existing one?

Comment (katmatt) [reply commercetools-models/src/main/java/io/sphere/sdk/products/PriceDraftBuilder.java]: Because the existing method has a different parameter type:
```
public PriceDraftBuilder custom(@Nullable final CustomFieldsDraft custom) {
```

heshamMassoud: ah ok, I got confused 😁

heshamMassoud: then does it make sense to have it in a separate method:
````java
  private static CustomFieldsDraft copyCustom(final CustomFields template) {
    return template == null ? null : CustomFieldsDraftBuilder.of(template).build();
  }
````

like it is in `CategoryDraftBuilder`, `AssetDraftBuilder`, etc?

katmatt: I totally forgot about them and yes, that's a very good idea =) And makes our code more consistent, thanks :1st_place_medal: I will update my PR 👍

PR #1548, opened 2017-09-26
making tests more precise

### LINKED ISSUE #1547: Create maven profile for vulnerability report
Since the `dependency-check-maven` plugin add latency to the building process (which become annoying in the case of repeatable builds 😢 )  i think it should be better to isolate it in a profile, hence it might be executed on demand.

acbeni:
Closes #1547

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ cache:
   directories:
   - $HOME/.m2
 script:
-  - mvn clean test verify javadoc:aggregate -P javadoc-jdk-8u121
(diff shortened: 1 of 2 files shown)


katmatt:
@acbeni Your change makes our build faster because you put the security check into a profile. Which is a good idea. But why does this change make our tests more precise?

acbeni:
@katmatt sorry didnt mean it makes the tests more precise but rather in case of rebuilding locally, skipping the report generation is a good thing to have 😕

acbeni:
@katmatt as agreed, the profile was added to travis build execution
thanks for pointing it out 🥇

PR #1549, opened 2017-09-26
Project update actions

acbeni:
Integration project update actions to our API


katmatt:
@acbeni Did you consider to generate the `ProjectUpdateCommand` with our code generator by adding the following annotations?
```
@ResourceInfo(pluralName = "projects", pathElement = "")
@HasUpdateCommand
```

acbeni:
@katmatt actually that was my first reflection but then the element `Project` should implement [`Versioned`](https://github.com/commercetools/commercetools-jvm-sdk/blob/master/commercetools-sdk-base/src/main/java/io/sphere/sdk/models/Versioned.java) unfortunatly the `Project` class lacks the attribute `id` wich wouldn't work for our model 😞

commercetools-models/src/main/java/io/sphere/sdk/projects/commands/ProjectUpdateCommand.java
```diff
@@ -0,0 +1,41 @@
...
+import io.sphere.sdk.projects.Project;
+
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Update commands for the {@link Project} entity
+ *{@include.example io.sphere.sdk.projects.commands.ProjectUpdateActionsIntegrationTest#execution()}
+ */
+@JsonDeserialize(as = ProjectUpdateCommandImpl.class)
```
katmatt: Is this annotation really necessary? We only serialize commands, but never derislize them.

commercetools-models/src/test/java/io/sphere/sdk/projects/commands/ProjectUpdateActionsIntegrationTest.java
```diff
@@ -0,0 +1,109 @@
...
+                ChangeName.of(PROJECT_NAME),
+                ChangeCurrencies.of(PROJECT_CURRENCIES),
+                ChangeCountries.of(PROJECT_COUNTRIES),
+                ChangeLanguages.of(PROJECT_LANGUAGES),
+                ChangeMessagesEnabled.of(PROJECT_MESSAGES_ENABLED)
+        ));
+
+        final Project updatedProject = client().executeBlocking(updateCommand);
+
+        softAssert(soft -> {
```
katmatt: I never used `softAssert`, but after reading the javadoc I really like it 👍

commercetools-models/src/test/java/io/sphere/sdk/projects/commands/ProjectUpdateActionsIntegrationTest.java
```diff
@@ -0,0 +1,109 @@
...
+    private static final String PROJECT_NAME = "TestProject";
+    private static final List<String> PROJECT_CURRENCIES = Arrays.asList(EUR.getCurrencyCode(), USD.getCurrencyCode());
+    private static final List<CountryCode> PROJECT_COUNTRIES = Arrays.asList(CountryCode.DE, CountryCode.US);
+    private static final List<Locale> PROJECT_LOCALES = Arrays.asList(Locale.GERMANY, new Locale("de", "AT"), Locale.ENGLISH);
+    private static final List<String> PROJECT_LANGUAGES = PROJECT_LOCALES.stream().map(Locale::toLanguageTag).collect(Collectors.toList());
+    private static final Boolean PROJECT_MESSAGES_ENABLED = true;
+
+    @BeforeClass
+    @AfterClass
+    public static void resetProject() {
```
katmatt: I like the idea to run this after and before the test 👍

commercetools-models/src/main/java/io/sphere/sdk/projects/commands/ProjectUpdateCommand.java
acbeni: 👍  good catch, actually its not necessary, so i would better eliminate it to avoid confusion.

PR #1554, opened 2017-09-27
Fixing CI fail regarding "trialUntil" parsing problem.

acbeni:


commercetools-models/src/main/java/io/sphere/sdk/projects/ProjectImpl.java
```diff
@@ -0,0 +1,20 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.neovisionaries.i18n.CountryCode;
+
+import javax.annotation.Nullable;
+import java.time.ZonedDateTime;
+import java.util.List;
+
+class ProjectImpl extends ProjectImplBase {
```
katmatt: I guess our code rule validation will report a missing `final` modifier here 😉

acbeni:
Yep making this class modifier makes sense 👍 
Good catch 😊

PR #1557, opened 2017-09-28
#1556: activate build profile avadoc-jdk-8u121 by default

### LINKED ISSUE #1556: Activate profile javadoc-jdk-8u121 by default
We should activate the `javadoc-jdk-8u121` in our maven build by default. The reasoning behind it is that more people are now using a new JDK 8 than an old one.

andrii-kovalenko-ct:
resolves #1556

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -547,11 +547,8 @@
         </profile>
         <profile>
             <id>javadoc-jdk-8u121</id>
-            <!--
-                This profile is currently only activated for the given JDK, but can be manually activated by passing
-                the -Pjavadoc-jdk-8u121 option to maven.
-            -->
             <activation>
+                <activeByDefault>true</activeByDefault>
                 <jdk>1.8.0_121</jdk>
             </activation>
             <properties>
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ cache:
   directories:
   - $HOME/.m2
 script:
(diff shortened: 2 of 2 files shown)


pom.xml
```diff
@@ -547,11 +547,8 @@
         </profile>
         <profile>
             <id>javadoc-jdk-8u121</id>
-            <!--
```
katmatt: Great that you removed this now unnecessary comment :1st_place_medal:

andrii-kovalenko-ct:
@katmatt 
>Only change request: Can you please remove the profile from the .travis file?

6e27484013

katmatt:
@andrii-kovalenko-ct I'm really sorry, but looks like we have to keep the profile in the `.travis` file because we added another one recently. And as soon as you specify one, the default ones are disabled. So please revert your last commit and then we can merge your PR.

andrii-kovalenko-ct:
@katmatt 
i had such a doubt, but i thought it is just a doubt :)
d764d8d5fd

andrii-kovalenko-ct:
don't you think we should activate both `javadoc-jdk-8u121` and `dependency-check` by default?

katmatt:
@andrii-kovalenko-ct No, `dependency-check` performs a security check of our dependencies and decreases the local build time a lot.

andrii-kovalenko-ct:
ok, got it

Issue #1527, opened 2017-09-12
Add convenience methods to generated builder classes


acbeni:
Avoid unnecessary extensions of the generated builder classes by adding convenience methods such as withers ...

in the case of collections find a way to initiate them when they are nullable in order to avoid null check later (which would allow to avoid bugs such as `plusFieldDefinitions causes NPE #1514`)

This might allow to remedy issue #1506.

acbeni:
@katmatt as discussed, it dosen't make a lot of sense to modify the code generator since it would be removed anyways later, a more simplistic idea would be: 

* Add the Class name when calling static method of mother classes in order to avoid problems with scala (for ex #1506)
* if possible, add a rule to check that we call mother static methods with the name of the mother class (if not possible, with a reasonable amount of effort, we can just rely on issues report to treat each case separately)
* See how we can translate the null collections semantic to empty collections in order to avoid issues such as #1514.

acbeni:
@katmatt issue update

> Add the Class name when calling static method of mother classes in order to avoid problems with scala (for ex #1506)

this doesn't seem quite the solution since through inheritance, the number of inherited static methods candidate for override become big and it adds a lot of noise to override them in the child classes.

>if possible, add a rule to check that we call mother static methods with the name of the mother class (if not possible, with a reasonable amount of effort, we can just rely on issues report to treat each case separately)

this seems not easy with current rule checker, maybe with the introduction of sonarQube or another style check framework we would be able to do so 

> See how we can translate the null collections semantic to empty collections in order to avoid issues such as #1514.

this has been successfully added through PR #1535 

@katmatt should we close this issue or keep it as a follow up for the scala compatibility issue?

katmatt:
I think we should close it and we should continue with #1506.

acbeni:
@katmatt sounds good for me 👍 
well keep only #1506

PR #1576, opened 2017-10-13
1572 draft builders setters

andrii-kovalenko-ct:
according to #1527:
  - update **all** `@ResourceDraftValue` drafts so they have _gettersForBuilder = true_
  - as a boy-scout approach: in some of those drafts re-order annotation properties so have the same order in all sources.


katmatt:
Created follow up issue #1580

PR #1578, opened 2017-10-16
Replace OWASP plugin with snyk vulnerability check.

### LINKED ISSUE #1574: Vulnerability check slows down build
Owasp makes our build slower than it already is, we should think about moving to OWASP.

acbeni:
fixes #1574

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,7 @@
 
 [![][travis img]][travis]
 [![][maven img]][maven]
+[![][snyk img]][snyk]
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -4,6 +4,7 @@
 
 [![][travis img]][travis]
 [![][maven img]][maven]
+[![][snyk img]][snyk]
```
katmatt: Cool, it even supports badges 👍

PR #1587, opened 2017-10-17
Order by order number

### LINKED ISSUE #1562: Get/Update/Delete order by order number
We should add support for getting, updating and deleting of an `Order` by it's order number. And we should take the chance to review our code generator, may be it makes sense to make it more generic with regards to the property that it uses in the generated code (currently we only support the more common `id` and `key`).

katmatt:
Closes #1562


acbeni:
But these tests fail even locally
`OrderDeleteCommandIntegrationTest`
`OrderUpdateCommandIntegrationTest`
`OrderQueryIntegrationTest`

katmatt:
@acbeni At least one test was successful 😉 Would be great if you take another look at my most recent changes 😄

acbeni:
@katmatt yep i saw it yesterday, i just restated the test that fails just to make sure it's just a temporary failure 😄

acbeni:
you are good to merge 👍

PR #1588, opened 2017-10-18
removing gettersForBuilder

### LINKED ISSUE #1580: Remove @ResourceDraftValue#gettersForBuilder
After we fixed #1572 we should remove the `@ResourceDraftValue#gettersForBuilder` and always generate getters for our `DraftBuilder` classes. This makes sure that our classes are more consistent in the future.

acbeni:
Closes #1580


commercetools-internal-processors/src/main/java/io/sphere/sdk/annotations/processors/generators/DraftBuilderGenerator.java
```diff
@@ -109,6 +104,13 @@ public TypeSpec generateType(final TypeElement resourceDraftValueType) {
         return draftBuilderBaseClass;
     }
 
+    private List<MethodSpec> createGettersForBuilders(final List<PropertyGenModel> properties){
```
katmatt: Nice that you extracted it to a separate method 👍

PR #1599, opened 2017-10-27
Sign in update product data

### LINKED ISSUE #1598: Add new updateProductData property to CustomerSignInCommand
We should add the new `updateProductData` boolean property to our CustomerSignInCommand.

katmatt:
Closes #1598


katmatt:
@acbeni Our JVM SDK currently doesn't support our `me` endpoints, mainly because that would require that we implement the missing oauth2 flow.

acbeni:
@katmatt ok i knew that we don't support the `me` endpoint but never knew why 🤔 
thanks @katmatt 😄 i think we are good to go 👍

PR #1621, opened 2017-11-28
in OSGi, client isn't always detected. by this the client is exposed …

acbeni:
Sometimes the client service passes undetected, this pr intend to correct this by exposing the client as an OSGi service.

### DIFF
--- a/commercetools-java-client-apache-async/pom.xml
+++ b/commercetools-java-client-apache-async/pom.xml
@@ -52,6 +52,11 @@
                         </Export-Package>
                         <Bundle-Version>${project.version}</Bundle-Version>
                         <Fragment-Host>${project.groupId}.commercetools-sdk-base</Fragment-Host>
+                        <Provide-Capability>
+                            osgi.serviceloader,
+                            osgi.serviceloader;
+                            osgi.serviceloader="io.sphere.sdk.client.SphereHttpClientFactory"
+                        </Provide-Capability>
                     </instructions>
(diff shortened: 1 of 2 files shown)


katmatt:
What do you mean with:
> Sometimes the client service passes undetected

And does this only happen in our tests?

acbeni:
@katmatt Actually when working with AEM, Markus had this message sometime 
```
27.11.2017 18:00:57.986 *ERROR* [OsgiInstallerImpl] com.commercetools.aem-connector.core [com.commercetools.aemconnector.core.CommercetoolsClientImpl(2998)] The activate method has thrown an exception (java.lang.NoClassDefFoundError: Could not initialize class io.sphere.sdk.client.SphereClientImpl)
java.lang.NoClassDefFoundError: Could not initialize class io.sphere.sdk.client.SphereClientImpl
    at io.sphere.sdk.client.SphereClient.of(SphereClient.java:46)
    at io.sphere.sdk.client.SphereClientFactory.createClient(SphereClientFactory.java:32)
    at com.commercetools.aemconnector.core.CommercetoolsClientImpl.activate(CommercetoolsClientImpl.java:78)
```

i figured out it was at the `SphereClientFactory` level and then it was due to the client service not detected,
so what i did is to expose the Apache client as a service, and make it as a required capability so the OSGi bundles won't start if no client is detected.

acbeni:
You're welcome 😄

PR #1625, opened 2017-11-29
Revert staged variant changes

acbeni:


acbeni:
closes #1626

commercetools-models/src/main/java/io/sphere/sdk/products/commands/updateactions/RevertStagedVariantChanges.java
```diff
@@ -0,0 +1,30 @@
+package io.sphere.sdk.products.commands.updateactions;
+
+import io.sphere.sdk.commands.UpdateActionImpl;
+import io.sphere.sdk.products.Product;
+
+/**
+ * evert changes of a variant, which were made to the staged version of a product and reset to the
```
katmatt: Looks like an "R" is missing here 😉

acbeni: good catch 👍

Issue #1636, opened 2017-12-04
Consider switshing to joda money 


acbeni:
Since we have had a nightmare with java money and it excessive use with class loader, OSGi ... maybe its time to find alternatives, i would like to see if joda-money would be a good fit or any other library.

acbeni:
Also `MoneyImpl` should no longer depend on `MonetaryAmount` because this kind of decoupling with a specific library make the sdk unable to change the implementation when needed.

PR #1637, opened 2017-12-05
Fixing java money bug osgi

acbeni:
this is a temporal fix for our client startup failure

related issue #1636


acbeni:
since moneta have a [milestone to release 1.2 on 12 december](https://github.com/JavaMoney/jsr354-ri/milestones) i think we should wait for the next release, to integrate it with this issue.

PR #1670, opened 2018-01-18
Add item draft missing attribute

acbeni:
Missing attribute in parcel draft add


commercetools-models/src/main/java/io/sphere/sdk/orders/ParcelDraft.java
```diff
@@ -1,23 +1,32 @@
 package io.sphere.sdk.orders;
 
+import io.sphere.sdk.annotations.FactoryMethod;
+import io.sphere.sdk.annotations.ResourceDraftValue;
+
 import javax.annotation.Nullable;
+import java.util.List;
+
 
+@ResourceDraftValue(factoryMethods = {@FactoryMethod(parameterNames = {})})
```
katmatt: Why do we need this annotation when our `ParcelDraftImpl` isn't generated yet?

acbeni: Yep i forgot to replace it by the generated class 😅 , now it's done!

PR #1687, opened 2018-01-26
Updating jackson version

### LINKED ISSUE #1601: Update Jackson dependencies
The latest jackson (2.9.2) has some problems with our SDK, updating to it was not a priority to us, but since now we are moving to AEM, there might be a case where it is imported as part of the transitive dependencies and these problems will surface again, so its time to evaluate updating to it.

acbeni:
Updating jackson version,

You might notice that i removed the constructor in `MessageSubscriptionPayloadImpl`, actually this is due to the fact that in the new jackson library you can't use `@JsonUnwrapped` combined with constructor.

closes #1601

### DIFF
--- a/commercetools-models/src/main/java/io/sphere/sdk/subscriptions/MessageSubscriptionPayloadImpl.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/subscriptions/MessageSubscriptionPayloadImpl.java
@@ -1,0 +1,36 @@
+package io.sphere.sdk.subscriptions;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonUnwrapped;
+import io.sphere.sdk.messages.Message;
+import io.sphere.sdk.models.Base;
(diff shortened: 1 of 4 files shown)


commercetools-models/src/main/java/io/sphere/sdk/subscriptions/MessageSubscriptionPayloadImpl.java
```diff
@@ -0,0 +1,36 @@
+package io.sphere.sdk.subscriptions;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonUnwrapped;
+import io.sphere.sdk.messages.Message;
+import io.sphere.sdk.models.Base;
+import io.sphere.sdk.models.Reference;
+
+final class MessageSubscriptionPayloadImpl<T> extends Base implements MessageSubscriptionPayload<T> {
```
katmatt: I'm curious why you had to add this class 😄

PR #1688, opened 2018-01-26
Solution info not working

### LINKED ISSUE #1685: SolutionInfo not working
Using the classloader of the SolutionInfoService prevents the SolutionInfo in my project to be found, as the loader doesn't know anything about my project's classes.

The change was introduced in release v1.24.0:
https://github.com/commercetools/commercetools-jvm-sdk/commit/f2f4ebd93d96ab82fa16aaf0e2791f08a99e7c1e#diff-9cbc3f81828a69f1a4b671c56b2ac151R13

acbeni:
closes #1685


acbeni:
i think the bug fix is transparent and not reported from any one external so no need to mention it

PR #1692, opened 2018-01-29
Solution Info bug

acbeni:
Solution info doesn't work quite well with Sunrise, this is a quick fix

### DIFF
--- a/commercetools-java-client-core/src/main/java/io/sphere/sdk/client/SolutionInfoService.java
+++ b/commercetools-java-client-core/src/main/java/io/sphere/sdk/client/SolutionInfoService.java
@@ -7,10 +7,9 @@ import java.util.*;
 //for architecture see https://docs.oracle.com/javase/tutorial/ext/basics/spi.html
 final class SolutionInfoService extends Base {
     private static SolutionInfoService instance;
-    private ServiceLoader<SolutionInfo> loader;
 
     private SolutionInfoService() {
-        loader = ServiceLoader.load(SolutionInfo.class);
+
     }
 
     public static synchronized SolutionInfoService getInstance() {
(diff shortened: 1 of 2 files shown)


osgi-support/commercetools-tests-fragment/pom.xml
```diff
@@ -110,7 +110,9 @@
                             filter:="(osgi.extender=osgi.serviceloader.processor)"
                         </Require-Capability>
                         <Provide-Capability>
-                            osgi.serviceloader
+                            osgi.serviceloader,
+                            osgi.serviceloader;
```
katmatt: This line is duplicated now, is that really intended here?

commercetools-java-client-core/src/main/java/io/sphere/sdk/client/SolutionInfoService.java
```diff
@@ -21,9 +20,20 @@ public static synchronized SolutionInfoService getInstance() {
     }
 
     public List<SolutionInfo> getSolutionInfos() {
-        final List<SolutionInfo> solutions = new LinkedList<>();
-        loader.iterator()
-                .forEachRemaining(solution -> solutions.add(solution));
+        List<SolutionInfo> solutions  ;
```
katmatt: We could have initialized the list here and this would have allowed us to make it final.

commercetools-java-client-core/src/main/java/io/sphere/sdk/client/SolutionInfoService.java
```diff
@@ -21,9 +20,20 @@ public static synchronized SolutionInfoService getInstance() {
...
+        // workaroud for play framework, since the overloaded version of ServiceLoader seems to cause some errors there, here we try bot until we get a working classloader
+        try{
+            loader = ServiceLoader.load(SolutionInfo.class);
+            solutions = new ArrayList<>();
+            loader.forEach(solutions::add);
+
+        }catch(Throwable e){
+            loader = ServiceLoader.load(SolutionInfo.class,SolutionInfo.class.getClassLoader());
+            solutions = new ArrayList<>();
+            loader.forEach(solutions::add);
```
katmatt: And we could have moved this `forEach` out of the try clause.

acbeni: Actually its during the insertion where the cast issue appears

commercetools-java-client-core/src/main/java/io/sphere/sdk/client/SolutionInfoService.java
acbeni: i am afraid that if an error happens during insertion we end up with a corrupt list 😢

osgi-support/commercetools-tests-fragment/pom.xml
acbeni: Yep this should be removed 👍

commercetools-java-client-core/src/main/java/io/sphere/sdk/client/SolutionInfoService.java
katmatt: I see 😄 and thanks for the explanation. I wasn't aware that the `load` method returns an `Iterator` 😉

PR #1721, opened 2018-02-28
Customer added to API Extensions PR

### LINKED ISSUE #1716: Cart API Extensions resourceTypeId naming conflict
Not sure if this is only a JVM SDK's naming issue or comes from the platform, but:

The `TriggerBuilder` class constructor is expecting something called "resourceTypeId" as first argument, although when creating an extension for cart the following is correct:
```
TriggerBuilder.of(Cart.referenceTypeId(), singletonList(TriggerType.CREATE)).build()
```
Whereas the following is not, because `Cart.resourceTypeId()` returns "order", instead of "cart" as the Trigger is expecting:
```
TriggerBuilder.of(Cart.resourceTypeId(), singletonList(TriggerType.CREATE)).build()
```

### LINKED ISSUE #1717: Customer added to API Extensions
We should add support for our backend feature "Customer added to API Extensions"

acbeni:
# Description

Closes #1717
Closes #1716

# Checklist

- [x] Update release Notes


commercetools-models/src/main/java/io/sphere/sdk/extensions/Trigger.java
```diff
@@ -13,7 +13,7 @@
 public interface Trigger {
 
 
-    String getResourceTypeId();
+    ExtendedResourceType getResourceTypeId();
```
katmatt: We should document this change in our release notes, just in case that someome used it already 😄

commercetools-models/src/main/java/io/sphere/sdk/extensions/ExtendedResourceType.java
```diff
@@ -0,0 +1,23 @@
+package io.sphere.sdk.extensions;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+
+public enum ExtendedResourceType {
```
katmatt: Why didn't you used our `SphereEnumeration` interface as a base? Then you wouldn't have to manually add the mapping logic here 😉

acbeni: Actually with sphere enumeration, you have to start with an uppercase, and i can't request that from the backend since for them its a resource id, its just further safety for us to make it an enum.

commercetools-models/src/main/java/io/sphere/sdk/extensions/Trigger.java
acbeni: Yep i ve done it 👍

lauraluiz: I feel like the name is a bit misleading. "ExtendedResourceType" sounds like the resource type has been extended with some properties. Simply "ExtensionResourceType" indicates that it's the resource type for extension. WDYT?

commercetools-models/src/main/java/io/sphere/sdk/extensions/ExtendedResourceType.java
katmatt: I see and then I'm fine with it 👍  May be we can add a comment that we can't use `SphereEnumeration` here together with your explanation?

commercetools-models/src/main/java/io/sphere/sdk/extensions/Trigger.java
katmatt: Sorry, looks like I didn't see it.

commercetools-models/src/main/java/io/sphere/sdk/extensions/ExtendedResourceType.java
acbeni: Sure, we definitely should keep track 👍

commercetools-models/src/main/java/io/sphere/sdk/extensions/Trigger.java
acbeni: @lauraluiz i too didn't find a good name, but yours sound good, i will change it 👍

acbeni:
@lauraluizjust checking, is it ok to move on with this?

PR #1743, opened 2018-04-10
Support for aws lambda extensions

### LINKED ISSUE #1741: Add AWS lambda destination to our api extension
Follow our backend API extension, see the corresponding PR.

acbeni:
# Description

Closes #1741

# Checklist

- [x] Update release Notes

### DIFF
--- a/commercetools-models/src/main/java/io/sphere/sdk/extensions/AWSLambdaDestination.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/extensions/AWSLambdaDestination.java
@@ -1,0 +1,18 @@
+package io.sphere.sdk.extensions;
+
+import com.fasterxml.jackson.annotation.JsonUnwrapped;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import io.sphere.sdk.annotations.FactoryMethod;
+import io.sphere.sdk.annotations.HasBuilder;
+import io.sphere.sdk.subscriptions.AwsCredentials;
+
+@JsonDeserialize(as = AWSLambdaDestinationImpl.class)
+@HasBuilder(factoryMethods = @FactoryMethod(parameterNames = {"arn","awsCredentials"}))
+public interface AWSLambdaDestination extends Destination{
+
+    String getArn();
+
+    @JsonUnwrapped
+    AwsCredentials getAwsCredentials();
+
+}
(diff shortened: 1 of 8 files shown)


commercetools-models/src/main/java/io/sphere/sdk/extensions/AWSLambdaDestination.java
```diff
@@ -0,0 +1,18 @@
...
+import io.sphere.sdk.annotations.HasBuilder;
+import io.sphere.sdk.subscriptions.AwsCredentials;
+
+@JsonDeserialize(as = AWSLambdaDestinationImpl.class)
+@HasBuilder(factoryMethods = @FactoryMethod(parameterNames = {"arn","awsCredentials"}))
+public interface AWSLambdaDestination extends Destination{
+
+    String getArn();
+
+    @JsonUnwrapped
```
katmatt: Nice that we could use this annotation here instead of having to deal with the properties 👍

commercetools-models/src/test/java/io/sphere/sdk/extensions/commands/AbstractExtensionIntegrationTest.java
```diff
@@ -14,35 +14,30 @@
...
 
-import static io.sphere.sdk.subscriptions.SubscriptionFixtures.assumeHasAzureFunctionUrl;
-import static io.sphere.sdk.subscriptions.SubscriptionFixtures.azureFunctionUrl;
 import static io.sphere.sdk.test.SphereTestUtils.asList;
 import static io.sphere.sdk.test.SphereTestUtils.randomKey;
 import static org.assertj.core.api.Assertions.assertThat;
 
 
-public class ExtensionUpdateCommandIntegrationTest extends IntegrationTest {
+public abstract class AbstractExtensionIntegrationTest extends IntegrationTest {
```
katmatt: I really like that you created an abstract test class here 👍

commercetools-models/src/main/java/io/sphere/sdk/extensions/AWSLambdaDestination.java
acbeni: Yep but we should consider including this in code generator, since the only reason i m not using it to generate the impl class here is that it create `JsonCreator`, which if combined with `JsonUnwrapped` causes jackson framework to throw an exception at runtime.

PR #1754, opened 2018-04-24
Async http client and spring boot 2 version clash

### LINKED ISSUE #1751: Async HTTP client and Spring Boot 2 version clash
Hi,

When we upgraded to Spring Boot 2, we ran into this error:
```
WARN  io.netty.util.concurre.DefaultPromise - An exception was thrown by org.asynchttpclient.netty.request.NettyRequestSender$1.operationComplete()
java.lang.NoSuchMethodError: io.netty.channel.DefaultChannelId.newInstance()Lio/netty/channel/DefaultChannelId;
```

As it turns out both [Spring Boot](https://github.com/spring-projects/spring-boot/blob/v2.0.1.RELEASE/spring-boot-project/spring-boot-dependencies/pom.xml#L132), and the [async-http-client](https://github.com/AsyncHttpClient/async-http-client/blob/2afae6007c76e32d6de3db41e0c297725aa1ff90/pom.xml#L379) (at least the [version present](https://github.com/commercetools/commercetools-jvm-sdk/blob/80070c46bac5c7ecbe252098690b4b5dd6316074/pom.xml#L85) in the Commercetools SDK) use the `netty.version` property.

This causes the Spring Boot property to override the async-http-client's Netty version, and since the version in Commercetools is a bit out of date it fails with the error linked above.

The workaround was to explicitly specify the older `netty.version` in our own `pom.xml`, and since nothing else uses Netty, this doesn't cause any issues (for us at least).

acbeni:
# Description

Closes #1751

# Checklist

- [x] Update release Notes

### DIFF
--- a/commercetools-java-client-internal-test/src/main/java/io/sphere/sdk/client/HttpClientAdapterTest.java
+++ b/commercetools-java-client-internal-test/src/main/java/io/sphere/sdk/client/HttpClientAdapterTest.java
@@ -105,6 +105,6 @@ public abstract class HttpClientAdapterTest {
         client.close();
         final String receivedHttpHeadersAndBody = future.join();
         LoggerFactory.getLogger(HttpClientAdapterTest.class).info(receivedHttpHeadersAndBody);
-        assertThat(receivedHttpHeadersAndBody).contains("Content-Length: " + length);
+        assertThat(receivedHttpHeadersAndBody).containsIgnoringCase("Content-Length: " + length);
     }
 }
--- a/commercetools-internal-docs/pom.xml
+++ b/commercetools-internal-docs/pom.xml
@@ -32,6 +32,12 @@
             <groupId>com.commercetools.sdk.jvm.core</groupId>
             <artifactId>commercetools-java-client</artifactId>
             <version>${project.version}</version>
+            <exclusions>
+                <exclusion>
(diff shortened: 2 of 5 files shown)


commercetools-java-client-internal-test/src/main/java/io/sphere/sdk/client/HttpClientAdapterTest.java
```diff
@@ -105,6 +105,6 @@ private void checkBodyRequest(final int port, final HttpRequestBody requestBody,
         client.close();
         final String receivedHttpHeadersAndBody = future.join();
         LoggerFactory.getLogger(HttpClientAdapterTest.class).info(receivedHttpHeadersAndBody);
-        assertThat(receivedHttpHeadersAndBody).contains("Content-Length: " + length);
+        assertThat(receivedHttpHeadersAndBody).containsIgnoringCase("Content-Length: " + length);
```
katmatt: That's interesting, I didn't expect that HTTP headers are case insensitive...

acbeni: Yep according to `RFC 2616` they are case insensitive and when i updated the client the test failed because they became lower case, so i had to ignore the case.

PR #1769, opened 2018-06-08
Price custom atribute bug fix

### LINKED ISSUE #1766: LineItemImportDraft should use PriceDraft instead of Price attributes
In `LineItemImportDraft` we use `Price` instead of `PriceDraft`, when a `Price` has custom fields this fields are deserialised with attribute `fieldsJsonMap` which is not recognised by the backend.

This issue should be fixed by replacing the `Price` with `PriceDraft` in `LineItemImportDraft`, however we should try to keep backward compatibility if possible.

acbeni:
# Description

Closes #1766

# Checklist

- [x] Update release Notes


commercetools-models/src/test/java/io/sphere/sdk/orders/commands/OrderImportCommandIntegrationTest.java
```diff
@@ -19,6 +19,8 @@
 import io.sphere.sdk.test.JsonNodeReferenceResolver;
 import io.sphere.sdk.test.SphereTestUtils;
 import io.sphere.sdk.types.CustomFields;
+import io.sphere.sdk.types.CustomFieldsDraft;
```
katmatt: Looks like these changes aren#t necessary. How about reverting them because they just add noise to our git history?

acbeni:
thanks @katmatt , the release notes has been updated, and the no necessary imports on OrderImportCommandIntegrationTest has been removed 👍

PR #1801, opened 2018-08-14
Add expansion to shipping method queries

### LINKED ISSUE #1800: Reference Expansion for shipping methods
We should add reference expansion for getting shipping methods by cart and by location.

katmatt:
# Description

Closes #1800 

# Checklist

- [x] Update release Notes

### DIFF
--- a/commercetools-models/src/main/java/io/sphere/sdk/shippingmethods/queries/ShippingMethodsByCartGetImpl.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/shippingmethods/queries/ShippingMethodsByCartGetImpl.java
@@ -8,7 +8,7 @@ import io.sphere.sdk.shippingmethods.expansion.ShippingMethodExpansionModel;
 
 import java.util.List;
 
-final class ShippingMethodsByCartGetImpl extends MetaModelGetDslImpl<List<ShippingMethod>, ShippingMethod, ShippingMethodsByCartGetImpl, ShippingMethodExpansionModel<ShippingMethod>> implements ShippingMethodsByCartGet {
+final class ShippingMethodsByCartGetImpl extends MetaModelGetDslImpl<List<ShippingMethod>, ShippingMethod, ShippingMethodsByCartGet, ShippingMethodExpansionModel<ShippingMethod>> implements ShippingMethodsByCartGet {
     ShippingMethodsByCartGetImpl(final String cartId) {
         super("?cartId=" + cartId, ShippingMethodEndpoint.ENDPOINT.withTypeReference(new TypeReference<List<ShippingMethod>>() {
             @Override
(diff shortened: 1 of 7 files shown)


commercetools-models/src/main/java/io/sphere/sdk/shippingmethods/queries/ShippingMethodsByCartGetImpl.java
```diff
@@ -1,24 +1,31 @@
...
 import io.sphere.sdk.queries.MetaModelGetDslBuilder;
 import io.sphere.sdk.queries.MetaModelGetDslImpl;
 import io.sphere.sdk.shippingmethods.ShippingMethod;
 import io.sphere.sdk.shippingmethods.expansion.ShippingMethodExpansionModel;
 
+import java.util.Collections;
 import java.util.List;
 
-final class ShippingMethodsByCartGetImpl extends MetaModelGetDslImpl<List<ShippingMethod>, ShippingMethod, ShippingMethodsByCartGetImpl, ShippingMethodExpansionModel<ShippingMethod>> implements ShippingMethodsByCartGet {
+final class ShippingMethodsByCartGetImpl extends MetaModelGetDslImpl<List<ShippingMethod>, ShippingMethod, ShippingMethodsByCartGet, ShippingMethodExpansionModel<ShippingMethod>> implements ShippingMethodsByCartGet {
```
lauraluiz: Uh that looked wrong before 😅 or was it actually necessary with the previous implementation?

commercetools-models/src/test/java/io/sphere/sdk/shippingmethods/queries/ShippingMethodsByLocationGetIntegrationTest.java
```diff
@@ -20,7 +20,7 @@
     public void execution() throws Exception {
         withShippingMethodForGermany(client(), shippingMethod -> {
             final SphereRequest<List<ShippingMethod>> sphereRequest =
-                    new VrapRequestDecorator<>(ShippingMethodsByLocationGet.of(CountryCode.DE), "response");
```
lauraluiz: Does it actually fail if the expansion path is not recognized? In a normal situation I know it fails silently, returning the query results as nothing happened, so it wouldn't fail. But as it's wrapped in this `VrapRequestDecorator` maybe it's checking it against the RAML specification?

commercetools-models/src/main/java/io/sphere/sdk/shippingmethods/queries/ShippingMethodsByCartGetImpl.java
katmatt: Good catch, I didn't notice that I changed this. But I agree that it looed wrong. And it doesn't seem to make any difference if you use `ShippingMethodsByCartGet` or `ShippingMethodsByCartGetImpl`. And this type parameter `<T>` is only used for expansion paths and I think that's why the "before" version worked.

commercetools-models/src/test/java/io/sphere/sdk/shippingmethods/queries/ShippingMethodsByLocationGetIntegrationTest.java
katmatt: This decorator is used to disable the response validation of our vrap proxy running on TeamCity and it then just pass through the response from our backend. And we have to disable the validation here because the same uri is used by multiple `GET` endpoints which return different types: `List<ShippingMethod>` vs `PagedQueryResult<ShippingMethod>`. And this can't be expressed in RAML and it's considered bad RESTful practise too 😉

lauraluiz: So that means that the test won't fail if `m.zones()` doesn't return the proper expansion path, right? If that's the case, then I would add an assertion checking that the returned shipping methods have zones expanded. The same for the test with Tax Category.

katmatt: The `VrapRequestDecorator` doesn't have any effect on the test if it's disabled. But you're right that the test won't fail when the expansion doesn't work 👍 I already added asserts to ensure that the expansion works.

PR #1802, opened 2018-08-21
Fix field name typo in cart update action SetShippingRateInput

lauraluiz:
# Description

There was a bug report that I couldn't confirm because the tests were not conclusive. I have improved them adding assertions in order to guarantee there's no bug present in the shipping rate inputs.

The output was that we had indeed a bug in the field name.

# Checklist

- [x] Update release Notes

### DIFF
--- a/commercetools-internal-docs/src/main/java/io/sphere/sdk/meta/ReleaseNotes.java
+++ b/commercetools-internal-docs/src/main/java/io/sphere/sdk/meta/ReleaseNotes.java
@@ -144,6 +144,7 @@ import java.util.function.Function;
  -->
  <h3 class=released-version id="v1_35_0">1.35.0 </h3>
  <ul>
+    <li class=fixed-in-release>Fixed field name in cart update action `SetShippingRateInput`, which was not properly working before</li>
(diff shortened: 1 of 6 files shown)


commercetools-models/src/test/java/io/sphere/sdk/carts/ShippingRateClassificationIntegrationTest.java
```diff
@@ -32,17 +32,15 @@
     @Test
     public void setClassificationShippingRateInput() {
 
-        final Set<LocalizedEnumValue> set = new HashSet();
+        final Set<LocalizedEnumValue> set = new HashSet<>();
```
katmatt: Good catch 👍

commercetools-internal-docs/src/main/java/io/sphere/sdk/meta/ReleaseNotes.java
```diff
@@ -144,6 +144,7 @@
  -->
  <h3 class=released-version id="v1_35_0">1.35.0 </h3>
  <ul>
+    <li class=fixed-in-release>Fixed field name in cart update action {@link io.sphere.sdk.carts.commands.updateactions.SetShippingRateInput}`, which was not properly working before</li>
```
katmatt: `fixed-in-release` is a very good classification: This is really a bug fix, because the update action never worked.

PR #1816, opened 2018-09-18
Set line item price

### LINKED ISSUE #1815: Set line item price
Follow up for backend enhancement.

acbeni:
# Description

Closes #1815 

# Checklist

- [x] Update release Notes

### DIFF
--- a/commercetools-models/src/main/java/io/sphere/sdk/orders/Order.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/orders/Order.java
@@ -224,8 +224,15 @@ public interface Order extends CartLike<Order> {
     @QueryModelHint(type = "DiscountCodeInfoCollectionQueryModel<Order>")
     List<DiscountCodeInfo> getDiscountCodes();
 
+    /**
+     * The refused gifts of this cart. Automatically filled when a line item with {@link LineItemMode#GIFT_LINE_ITEM}
+     * is removed from this order.
+     *
+     * @return the refused gifts of this cart
+     */
     @Nullable
-    List<CartDiscount> getRefusedGifts();
+    @IgnoreInQueryModel
+    List<Reference<CartDiscount>> getRefusedGifts();
 
 
     @Nullable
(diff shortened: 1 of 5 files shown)


commercetools-models/src/main/java/io/sphere/sdk/orders/Order.java
```diff
@@ -223,6 +224,17 @@ public String toString() {
...
 
+    /**
+     * The refused gifts of this cart. Automatically filled when a line item with {@link LineItemMode#GIFT_LINE_ITEM}
+     * is removed from this order.
+     *
+     * @return the refused gifts of this cart
+     */
+    @Nullable
+    @IgnoreInQueryModel
+    List<Reference<CartDiscount>> getRefusedGifts();
```
katmatt: I would suggest to add a test here. Just to make sure that the name is correct. What do you think about it?

acbeni: didn't quit understood, here i didn't add a test because making `getRefusedGifts` fields non null needs order edit interaction which wasn't included in the SDK

acbeni:
@lauraluiz a pointer to the issue

commercetools-models/src/main/java/io/sphere/sdk/orders/Order.java
lauraluiz: I see. It is quite problematic for the SDK not to be able to modify a value, it limits tests a lot. Do you have any other solution for these situations? Only solution (without implementing OrderEdit) I see is to test it manually once, to at least have the peace of mind that this actually works. That might be an overkill, I know. But merging code that no one has check it actually works, that sounds like risky 😨

acbeni: yep... but i cant really think of a solution 😢 , maybe i can make a deserialization from json, then i know if my format corresponds to the one specified t the ticket it should work 🤔 i thimnk that should b good enough

acbeni:
Added test case.

PR #1837, opened 2018-10-24
#1810 resolve http client conflict

### LINKED ISSUE #1810: Conflicts with updated major version of asynchttpclient
In this PR https://github.com/commercetools/commercetools-jvm-sdk/pull/1754/files the asynchttpclient was updated to a major version. But this usually carries issues, that's why the JVM SDK supports each major version separately (1.8, 1.9, 2.0). Currently I'm facing issues updating the JVM SDK dependency in SUNRISE because of this.

Additionally, you can see that, in order to update the major version of the asynchttpclient, the maven property holding the v2.0 of the library now holds 2.4.x version, thus not matching its name:
https://github.com/commercetools/commercetools-jvm-sdk/blob/master/pom.xml#L85

I think we should stick to having separate client versions as before (as @schleichardt designed it), unless we are 100% sure that this will not affect other projects.

acbeni:
# Description

Closes #1810

# Checklist

- [x] Update release Notes


pom.xml
```diff
@@ -82,7 +84,8 @@
         <gson.version>2.5</gson.version>
         <asynchttpclient18.version>1.8.16</asynchttpclient18.version>
         <asynchttpclient19.version>1.9.33</asynchttpclient19.version>
-        <asynchttpclient20.version>2.4.5</asynchttpclient20.version>
+        <asynchttpclient20.version>2.0.38</asynchttpclient20.version>
+        <asynchttpclient25.version>2.4.5</asynchttpclient25.version>
```
lauraluiz: Given that it's 2.4.5, shouldn't it be called `asynchttpclient24` instead?

sdk-http-ahc-2_5/src/main/java/io/sphere/sdk/http/AsyncHttpClientAdapter.java
```diff
@@ -0,0 +1,13 @@
+package io.sphere.sdk.http;
+
+import org.asynchttpclient.AsyncHttpClient;
+
+/**
+ * Adapter to use {@code AsyncHttpClient} (version 2.0.x) as {@code HttpClient}.
```
lauraluiz:
```suggestion
 * Adapter to use {@code AsyncHttpClient} (version 2.4.x) as {@code HttpClient}.
```

acbeni: good that you pointed it out , i was referring to 2.5.4, i will fix the docs

PR #1846, opened 2018-11-19
Add zone key

### LINKED ISSUE #1845: Adding  zone.key attribute and change the concerned entities
changes following some back-end changes

acbeni:
initial commit

# Description

Closes #1845

# Checklist

- [x] Update release Notes

### DIFF
--- a/commercetools-models/src/main/java/io/sphere/sdk/shippingmethods/commands/updateactions/RemoveZone.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/shippingmethods/commands/updateactions/RemoveZone.java
@@ -3,6 +3,7 @@ package io.sphere.sdk.shippingmethods.commands.updateactions;
 import io.sphere.sdk.commands.UpdateActionImpl;
 import io.sphere.sdk.models.Reference;
 import io.sphere.sdk.models.Referenceable;
+import io.sphere.sdk.models.ResourceIdentifier;
 import io.sphere.sdk.shippingmethods.ShippingMethod;
 import io.sphere.sdk.zones.Zone;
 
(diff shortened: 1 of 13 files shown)


commercetools-models/src/main/java/io/sphere/sdk/zones/Zone.java
```diff
@@ -59,14 +60,22 @@
...
-@HasDeleteCommand
+@HasUpdateCommand(javadocSummary = "Updates a zone.", updateWith = "key")
+@HasDeleteCommand(deleteWith = {"key","id"})
 @HasQueryModel
 public interface Zone extends Resource<Zone> {
 
     String getName();
 
+    /**
+     * @return Unique identifier for a zone. Must be unique across a project.
```
katmatt: Nice javadoc 👍

commercetools-models/src/main/java/io/sphere/sdk/shippingmethods/commands/updateactions/RemoveZone.java
```diff
@@ -14,18 +15,18 @@
...
+    private final ResourceIdentifier<Zone> zone;
 
-    private RemoveZone(final Reference<Zone> zone) {
+    private RemoveZone(final ResourceIdentifier<Zone> zone) {
         super("removeZone");
         this.zone = zone;
     }
 
-    public Reference<Zone> getZone() {
+    public ResourceIdentifier<Zone> getZone() {
```
katmatt: I would suggest to mention this change in the release notes too. May be like telling our users that we changed all update actions which referenced a zone to use a `ResourceIdentifier<Zone>` instead of a `Reference<Zone>`. What do yo think about it?

acbeni: totally agree, would add it.

PR #1850, opened 2018-11-28
update following rules

### LINKED ISSUE #1848: GiftLineItemCartDiscountValue takes References when it should take ResourceIdentifiers
#### Description
The `GiftLineItemCartDiscountValue` in the JVM SDK currently only takes `Reference`s for `product`, `supplyChannel` and `distributionChannel` when it actually should take `ResourceIdentifier`s for all, since our [API already allows that](https://docs.commercetools.com/http-api-projects-cartDiscounts.html#gift-line-item). 

Here is the class in the JVM SDK:
https://github.com/commercetools/commercetools-jvm-sdk/blob/2080b5b5d827ce8760d2c6937b1220c415dff920/commercetools-models/src/main/java/io/sphere/sdk/cartdiscounts/GiftLineItemCartDiscountValue.java#L26-L34

This makes it cumbersome for users who want to create their `CartDiscount`s with keys right away for the product and channels.

acbeni:
# Description

Closes #1848 

# Checklist

- [x] Update release Notes


heshamMassoud:
Thanks and looks good! It would be nice to have unit tests for it 😉

PR #1871, opened 2019-01-18
#1868 set of references expansion

### LINKED ISSUE #1868: Attribute of type Set of expanded references ignores the expansion `obj`
#### Current Behaviour 
When I create an attribute of a `Set` of `Product` references type where the product references are expanded, somehow the `obj` expansion is ignored:

For example:
```java
final Product product = readObjectFromResource(PRODUCT_KEY_1_RESOURCE_PATH, Product.class);

final Reference<Product> expandedReference =
Reference.ofResourceTypeIdAndIdAndObj(Product.referenceTypeId(), UUID.randomUUID().toString(), product);

final HashSet<Reference<Product>> references = new HashSet<>();
references.add(expandedReference);

final Attribute attribute = Attribute.of("attrName", AttributeAccess.ofProductReferenceSet(), references);
```

The created `attribute` in the end looks like this:
```java
AttributeImpl[name=attrName,value=[{"typeId":"product","id":"07421bdc-cfeb-458c-8810-e62525fe291d"}]]
```

The expanded `obj` gets completely ignored.

#### Expected Behaviour

The `obj` shouldn't be ignored.

alex-ct:
# Description

Closes #1868 

# Checklist

- [x] Update release Notes

### DIFF
--- a/commercetools-models/src/main/java/io/sphere/sdk/products/attributes/Attribute.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/products/attributes/Attribute.java
@@ -2,7 +2,10 @@ package io.sphere.sdk.products.attributes;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.sphere.sdk.json.SphereJsonUtils;
 import io.sphere.sdk.models.EnumValue;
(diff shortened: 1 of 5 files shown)


commercetools-models/src/main/java/io/sphere/sdk/products/attributes/ReferenceInternalSerializer.java
```diff
@@ -0,0 +1,29 @@
...
+        super(Reference.class,true);
+    }
+
+    @Override
+    public Class<Reference> handledType() {
+        return super.handledType();
+    }
+
+    @Override
+    public void serialize(Reference reference, JsonGenerator jsonGenerator, SerializerProvider serializerProvider) throws IOException {
```
acbeni: add `final` modifier for attributes here

commercetools-models/src/main/java/io/sphere/sdk/products/attributes/Attribute.java
```diff
@@ -36,16 +39,14 @@ static Attribute of(final String name, final JsonNode jsonNode) {
...
-        //here is not the attributeMapper used to keep LocalizedEnum values which
-        //are transformed to just the key so the attribute could not be read anymore
-        final JsonNode jsonNode = SphereJsonUtils.toJsonNode(value);
-        if (value instanceof Reference && jsonNode instanceof ObjectNode) {
-            final Reference<?> reference = (Reference<?>) value;
-            if (reference.getObj() != null) {
-                ((ObjectNode) jsonNode).replace("obj", SphereJsonUtils.toJsonNode(reference.getObj()));
-            }
-        }
+        final ObjectMapper mapper = SphereJsonUtils.newObjectMapper();
```
katmatt: Here we could try to store the `ObjectMapper` in a static field - preferably in a package protected helper class.
This way we would avoid that we recreate it again and again. And as far as I can see this should be thread safe too. What do you think about it?

acbeni: sounds good to me too @katmatt i had also performance concern, but had no idea how to fix it 👍

commercetools-models/src/main/java/io/sphere/sdk/products/attributes/ReferenceInternalObjectMapper.java
```diff
@@ -0,0 +1,21 @@
+package io.sphere.sdk.products.attributes;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import io.sphere.sdk.json.SphereJsonUtils;
+
+class ReferenceInternalObjectMapper {
```
acbeni: make this class final

PR #1891, opened 2019-02-19
#1614 get current product

### LINKED ISSUE #1614: Unexpected behaviour when getting Current Product?
I have to get the Current product even if it is unpublished.
As a platform user I expect that I will just fetch the product from Products endpoint and then

```java
product.getMasterData().getCurrent()
```

where `product` is an object created from `io.sphere.sdk.products.Product`.

However SDK decides instead of me that I shouldn't get the current product if it is unpublished 

https://github.com/commercetools/commercetools-jvm-sdk/blob/8055d7a99ccd9126a6431151ccee061971850282/commercetools-models/src/main/java/io/sphere/sdk/products/ProductCatalogDataImpl.java#L23-L25

I think SDK should return exactly what I fetched from the platform and not add some custom logic layer on that.

alex-ct:
# Description

Closes #1614 

# Checklist

- [x] Update release Notes

### DIFF
--- a/commercetools-internal-docs/src/main/java/io/sphere/sdk/meta/ReleaseNotes.java
+++ b/commercetools-internal-docs/src/main/java/io/sphere/sdk/meta/ReleaseNotes.java
@@ -164,6 +164,7 @@ import java.util.function.Function;
     <li class=new-in-release>Added {@link io.sphere.sdk.orderedits.OrderEdit} update and staged actions</li>
     <li class=new-in-release>Added {@link io.sphere.sdk.orderedits.commands.OrderEditApplyCommand}</li>
     <li class=new-in-release>Added new {@link io.sphere.sdk.orderedits.OrderEdit} message types</li>
+    <li class=new-in-release><Added {@link ProductCatalogData#getCurrentUnsafe()} method/li>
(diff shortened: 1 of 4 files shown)


commercetools-internal-docs/src/main/java/io/sphere/sdk/meta/ReleaseNotes.java
```diff
@@ -164,6 +164,7 @@
     <li class=new-in-release>Added {@link io.sphere.sdk.orderedits.OrderEdit} update and staged actions</li>
     <li class=new-in-release>Added {@link io.sphere.sdk.orderedits.commands.OrderEditApplyCommand}</li>
     <li class=new-in-release>Added new {@link io.sphere.sdk.orderedits.OrderEdit} message types</li>
+    <li class=new-in-release><Added {@link ProductCatalogData#getCurrentUnsafe()} method/li>
```
acbeni: the closing `</li>` is missing not correct.

alex-ct: Good catch !

PR #1905, opened 2019-04-09
Extract external oauth2 test

katmatt:
# Description

This extracts our flaky integration test for setting the external oauth property on the project level to a separate test. This test is only executed if an api client for a new project is set via environment variables with `JVM_SDK_IT_OAUTH` prefix. I added these environment variables to our travis.


alex-ct:
Look good to me 👍

PR #1911, opened 2019-04-17
#1881 add stores

### LINKED ISSUE #1881: Introduce Stores in Beta
introducing store backend feature to the jvm sdk.

alex-ct:
# Description

Closes #1881 

# Checklist

- [x] Update release Notes


commercetools-sdk-base/src/main/java/io/sphere/sdk/client/SphereAuthConfig.java
```diff
@@ -37,6 +39,10 @@
         return DEFAULT_SCOPES;
     }
 
+    default List<String> getRawScopes() {
```
katmatt: Here we should add some javadoc to describe the difference between this method and the `getScopes()` method.

alex-ct:
The answer from support was that the new scopes are only supported for the graphql. I will merge the PR as the test for the scopes is ignored. When they add the support, we can simply uncomment the ignore annotation.

acbeni:
actually if they are supported for graphql then i would suggest to totally remove them, since i don't think by the answer that there is a plan to port these features to the rest API.

PR #1958, opened 2019-07-19
1957 Extend WithKey interface in Category, CategoryDraft, ProductLike and ProductDraft.

### LINKED ISSUE #1957: Category and Product resources should extend WithKey interface
`Category` and `Product` resources should extend the `WithKey` interface in order to get accessibility on the method `getKey` when declaring a class with generic parameters that extends the `WithKey` interface.

heshamMassoud:
# Description
- Extends WithKey interface in Category, CategoryDraft, ProductLike and ProductDraft.
Closes #1957 

# Checklist

- [x] Update release Notes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: java
 jdk:
-  - oraclejdk8
+  - openjdk8
 cache:
   directories:
   - $HOME/.m2
--- a/commercetools-models/src/test/java/io/sphere/sdk/categories/CategoryDraftTest.java
+++ b/commercetools-models/src/test/java/io/sphere/sdk/categories/CategoryDraftTest.java
@@ -1,0 +1,27 @@
+package io.sphere.sdk.categories;
+
+import io.sphere.sdk.models.WithKey;
+import org.junit.Test;
+
+import javax.annotation.Nonnull;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class CategoryDraftTest {
+
+    @Test
+    public void categoryDraftImplementsWithKey() {
+        final CategoryDraft category = mock(CategoryDraft.class);
+        when(category.getKey()).thenReturn("foo");
+
+        final String key = getKey(category);
+
+        assertThat(key).isEqualTo(category.getKey());
+    }
+
+    private <T extends WithKey> String getKey(@Nonnull final T resource) {
+        return resource.getKey();
+    }
+}
--- a/commercetools-models/src/test/java/io/sphere/sdk/categories/CategoryDraftTest.java
+++ b/commercetools-models/src/test/java/io/sphere/sdk/categories/CategoryDraftTest.java
@@ -21,7 +21,7 @@ public class CategoryDraftTest {
         assertThat(key).isEqualTo(category.getKey());
     }
 
-    private <T extends WithKey> String getKey(@Nonnull final T resource) {
+    public <T extends WithKey> String getKey(@Nonnull final T resource) {
         return resource.getKey();
     }
 }
--- a/commercetools-models/src/test/java/io/sphere/sdk/categories/CategoryDraftTest.java
+++ b/commercetools-models/src/test/java/io/sphere/sdk/categories/CategoryDraftTest.java
@@ -1,5 +1,6 @@
 package io.sphere.sdk.categories;
 
+import io.sphere.sdk.annotations.NotOSGiCompatible;
 import io.sphere.sdk.models.WithKey;
 import org.junit.Test;
 
@@ -9,6 +10,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
+@NotOSGiCompatible
 public class CategoryDraftTest {
 
     @Test
--- a/commercetools-models/src/test/java/io/sphere/sdk/categories/CategoryDraftTest.java
+++ b/commercetools-models/src/test/java/io/sphere/sdk/categories/CategoryDraftTest.java
@@ -23,7 +23,7 @@ public class CategoryDraftTest {
         assertThat(key).isEqualTo(category.getKey());
     }
 
-    public <T extends WithKey> String getKey(@Nonnull final T resource) {
+    private <T extends WithKey> String getKey(@Nonnull final T resource) {
         return resource.getKey();
     }
 }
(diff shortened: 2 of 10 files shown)


acbeni:
Thank you @heshamMassoud for the contribution, it makes a lot of sense in our sdk, i ll just try to fix our build problem in your branch in order to be able to merge it.

commercetools-models/src/test/java/io/sphere/sdk/categories/CategoryDraftTest.java
```diff
@@ -21,7 +21,7 @@ public void categoryDraftImplementsWithKey() {
         assertThat(key).isEqualTo(category.getKey());
     }
 
-    private <T extends WithKey> String getKey(@Nonnull final T resource) {
+    public <T extends WithKey> String getKey(@Nonnull final T resource) {
```
heshamMassoud: @acbeni I am wondering why would we change this `private` utils to `public`? 🤔

acbeni: it's for the way we package our tests, actually we use  a [custom plugin](https://github.com/commercetools/commercetools-jvm-sdk/tree/master/osgi-support/mirror-subclass-plugin) to subclass all the tests and this is done in order to add some annotations that are specific to [PaxExam](https://ops4j1.jira.com/wiki/spaces/PAXEXAM4/overview) via byte code manipulation, these annotations make sure that the test is run via pax exam runner as in the [example](https://ops4j1.jira.com/wiki/spaces/PAXEXAM3/pages/27295748/Getting+Started+with+OSGi+Tests), all these extensions/subclassing operations makes it difficult to work with private methods, and this is why i switch methods (in test) to public, it's not a big deal since they are not shipped to our clients, and in general it gives us one less aspect to take care of.

heshamMassoud: hmm, thanks for this explanation @acbeni  Although, this is confusing for me given the error messages are:
```
categoryDraftImplementsWithKey(io.sphere.sdk.categories.___CategoryDraftTest)  Time elapsed: 0.04 sec  <<< ERROR!
java.lang.ClassNotFoundException: org.mockito.Mockito not found

productDraftImplementsWithKey(io.sphere.sdk.products.___ProductDraftTest)  Time elapsed: 0.026 sec  <<< ERROR!
java.lang.ClassNotFoundException: org.mockito.Mockito not found

productLikeImplementsWithKey(io.sphere.sdk.products.___ProductLikeTest)  Time elapsed: 0.014 sec  <<< ERROR!
java.lang.ClassNotFoundException: org.mockito.Mockito not found 
```

acbeni: yep figured out i need to rectify the configuration  [here](https://github.com/commercetools/commercetools-jvm-sdk/blob/66a3ce79b10b97d6deef1a19a8a3586db015fac2/osgi-support/sdk-osgi-test-campaign/src/main/java/io/sphere/sdk/test/JvmSdkConfigurationFactory.java)  but i opted for the easier way c7e02e7 as not to run them in the osgi container.

acbeni: not the best solution but its a fast one, and anyways what matters is that the integration tests run fine in OSGi which they still do.

heshamMassoud: also this [one](https://github.com/commercetools/commercetools-jvm-sdk/pull/1958/files#diff-e5c9294211de558467e60b73075779faR45) is still `private` and not failing.

Is it not because of the `Mockito#mock` usage? (These three failing tests are using it)

acbeni: You are right, i just reviewed the logs and it seems so, its the Mockito dependency that shows in the error, i would expect `@NotOSGiCompatible` to fix the problem, since it prevents this test to be run in an OSGi container, if not i ll investigate the problem deeper

heshamMassoud: > yep figured out i need to rectify the configuration here 

From the link you sent, seems like mockito dependency is there 🤔: https://github.com/commercetools/commercetools-jvm-sdk/blob/66a3ce79b10b97d6deef1a19a8a3586db015fac2/osgi-support/sdk-osgi-test-campaign/src/main/java/io/sphere/sdk/test/JvmSdkConfigurationFactory.java#L75


> but i opted for the easier way c7e02e7 as not to run them in the osgi container.

hmm, but then it's weird that this test [`CategoryTest`](https://github.com/commercetools/commercetools-jvm-sdk/pull/1958/files#diff-e5c9294211de558467e60b73075779fa) runs successfully without this annotation..

> but i opted for the easier way c7e02e7 as not to run them in the osgi container.

shall we make the `private` again then?

acbeni: >hmm, but then it's weird that this test CategoryTest runs successfully without this annotation.

`CategoryTest`  actually doesn't have a dependency on Mockito as an import, it rather deserializes resources from json, this is why it runs fine

> From the link you sent, seems like mockito dependency is there

yep but in OSGi you have to import it in the `commercetools-sdk-base`'s manifest, which is not done since it just a test depedency, i m sure there is a solution for that, but i still opt for the easy `@NotOSGiCompatible` to unblock you faster.

heshamMassoud: > CategoryTest actually doesn't have a dependency on Mockito as an import, it rather deserializes resources from json, this is why it runs fine

True, right :) 

> yep but in OSGi you have to import it in the commercetools-sdk-base's manifest, which is not done since it just a test depedency, i m sure there is a solution for that, but i still opt for the easy @NotOSGiCompatible to unblock you faster.

Ok, I am not so into this topic. But thanks for the explanation. 

> definitely, i ll revert my changes
👍

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: java
 jdk:
-  - oraclejdk8
+  - openjdk8
```
acbeni: Just asking for an extra opinion here, @katmatt we had this problem https://travis-ci.community/t/error-installing-oraclejdk8-expected-feature-release-number-in-range-of-9-to-14-but-got-8/3766/4 and had to switch to openjdk to fix it, is that fine with you ?

katmatt: That's fine with me. We anyway have to discuss how we want to proceed with supporting multiple JDK versions and distributions 😄

acbeni: cool, we ll this would solve our problems until we see we re the discussion in travis resolves.

acbeni:
@heshamMassoud i think we re good to merge this one 👍

heshamMassoud:
@acbeni @katmatt (not entirely sure if this is the correct process) but:
- Added a new entry for the release notes: 69cba90 
- Fixed a typo in the previous release notes : a60b7fd

acbeni:
Yep this is good @heshamMassoud 
If fine with @katmatt @alex-ct , we ll make a release at the end of week to unblock you.

PR #1997, opened 2019-12-18
ParcelDraft allow setting the parcel items

FL-K:
# Description
We want to be able to set the list of delivery items for the parcel.
Right now it is not possible to specify this in the `ParcelDraft` as none of the builder methods accept argument of type  List<DeliveryItem>.

We also tried doing this process in two steps:
First creating delivery for the order
Then updating the order with `AddParcelToDelivery` action
This did not work also because the last argument of type List<DeliveryItem> is ignored and parcel is shown in merchant center always with 0 items.
```
AddParcelToDelivery.of(
             delivery,
             parcel,
             deliveryItems // this is ignored, the items here are not added to the parcel
);
```

### DIFF
--- a/commercetools-models/src/main/java/io/sphere/sdk/orders/ParcelDraft.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/orders/ParcelDraft.java
@@ -22,11 +22,19 @@ public interface ParcelDraft {
         return new ParcelDraftDsl(null ,measurements, trackingData);
     }
 
+    static ParcelDraft of(final ParcelMeasurements measurements, final TrackingData trackingData, final List<DeliveryItem> items) {
+        return new ParcelDraftDsl(items ,measurements, trackingData);
+    }
+
     static ParcelDraft of(final ParcelMeasurements measurements) {
         return new ParcelDraftDsl(null, measurements, null);
     }
 
     static ParcelDraft of(final TrackingData trackingData) {
         return new ParcelDraftDsl(null, null, trackingData);
+    }
+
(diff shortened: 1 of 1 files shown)


katmatt:
@FL-K Thanks for your PR. We still support builder methods on our draft interfaces, but we recommend that you instead use our more powerful `*DraftBuilder` classes. They are much easier to use and reduce the complexity of our builder methods. Imagine we would have to support builder methods with all combination of properties, that would add a lot of complexity to our SDK. 

Here you can find the javadoc for our [ParcelDraftBuilder](http://commercetools.github.io/commercetools-jvm-sdk/apidocs/io/sphere/sdk/orders/ParcelDraftBuilder.html) 

Please let us know if it works for you and if it works for you, we would close this PR.

FL-K:
> @FL-K Thanks for your PR. We still support builder methods on our draft interfaces, but we recommend that you instead use our more powerful `*DraftBuilder` classes. They are much easier to use and reduce the complexity of our builder methods. Imagine we would have to support builder methods with all combination of properties, that would add a lot of complexity to our SDK.
> 
> Here you can find the javadoc for our [ParcelDraftBuilder](http://commercetools.github.io/commercetools-jvm-sdk/apidocs/io/sphere/sdk/orders/ParcelDraftBuilder.html)
> 
> Please let us know if it works for you and if it works for you, we would close this PR.

Thanks for the reply, yes ParcelDraftBuilder solved the issue for us.
Used the ParcelDraftBuilder with ParcelDraft then setting the items.

I am closing this PR

PR #2021, opened 2020-06-17
Add missing includedInPrice property to ExternalTaxRateDraft.

### LINKED ISSUE #2020: Add missing includedInPrice property to ExternalTaxRateDraft.

katmatt:
# Description

Closes #2020

# Checklist

- [x] Update release Notes

### DIFF
--- a/commercetools-models/src/main/java/io/sphere/sdk/taxcategories/ExternalTaxRateDraft.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/taxcategories/ExternalTaxRateDraft.java
@@ -22,6 +22,8 @@ public interface ExternalTaxRateDraft {
     @Nullable
     String getState();
 
+    Boolean isIncludedInPrice();
+
     /**
      * For countries (e.g. the US) where the total tax is a combination of multiple taxes (e.g. state and local taxes).
      * @return sub rates or null
--- a/commercetools-models/src/main/java/io/sphere/sdk/taxcategories/ExternalTaxRateDraft.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/taxcategories/ExternalTaxRateDraft.java
@@ -22,6 +22,7 @@ public interface ExternalTaxRateDraft {
     @Nullable
     String getState();
 
+    @Nullable
     Boolean isIncludedInPrice();
 
     /**
--- a/commercetools-models/src/main/java/io/sphere/sdk/taxcategories/ExternalTaxRateDraftBuilder.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/taxcategories/ExternalTaxRateDraftBuilder.java
@@ -16,6 +16,7 @@ public final class ExternalTaxRateDraftBuilder extends Base implements Builder<E
     private String state;
     @Nullable
     private List<SubRate> subRates;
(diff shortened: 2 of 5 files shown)


ashishhk:
@jenschude Could you please review this one ?

commercetools-models/src/main/java/io/sphere/sdk/taxcategories/ExternalTaxRateDraftBuilder.java
```diff
@@ -16,6 +16,7 @@
     private String state;
     @Nullable
     private List<SubRate> subRates;
+    private Boolean includedInPrice;
```
jenschude, line 19: Shouldn't this also declared nullable or set to a default value? Documentation says defaults to false

commercetools-models/src/main/java/io/sphere/sdk/taxcategories/ExternalTaxRateDraft.java
```diff
@@ -22,6 +22,8 @@
     @Nullable
     String getState();
 
+    Boolean isIncludedInPrice();
```
jenschude, line 25: Should be nullable or default should be set.

commercetools-models/src/test/java/io/sphere/sdk/carts/ExternalTaxRatesIntegrationTest.java
```diff
@@ -145,14 +145,15 @@ public void setLineItemTaxRate() {
                 final String taxRateName = "special tax";
                 final double taxRate = 0.20;
                 final ExternalTaxRateDraft externalTaxRate =
-                        ExternalTaxRateDraftBuilder.ofAmount(taxRate, taxRateName, DE).build();
+                        ExternalTaxRateDraftBuilder.ofAmount(taxRate, taxRateName, DE).includedInPrice(false).build();
```
jenschude, line 148: I would have added this as an additional test to check as this flag can have a default value

commercetools-models/src/main/java/io/sphere/sdk/taxcategories/ExternalTaxRateDraftBuilder.java
katmatt: you're right with regards to the docs. But if the field is set to `null`, then the backend will set it to the default value `false`. And I don't think it makes sense to replicate this in our JVM SDK. Hope that makes sense to you too!

commercetools-models/src/main/java/io/sphere/sdk/taxcategories/ExternalTaxRateDraft.java
katmatt: See comment above and it shouldn't be `@Nullable`, because we only use that annotation for optional fields.

commercetools-models/src/test/java/io/sphere/sdk/carts/ExternalTaxRatesIntegrationTest.java
katmatt: Yes, ideally I would have spent more time on a separate test. But we try to spent just the minimal time on the JVM SDK and that's why I didn't create a separate test.

Issue #2026, opened 2020-06-29
support localeProjection in Product Projection Search


barbara79:
https://jira.commercetools.com/browse/SUPPORT-9178

PR #2027, opened 2020-07-01
Support locale product projection

barbara79:
# Description

Closes https://github.com/commercetools/commercetools-jvm-sdk/issues/2026

# Checklist

- [x] Update release Notes

### DIFF
--- a/commercetools-models/src/main/java/io/sphere/sdk/products/search/PriceSelection.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/products/search/PriceSelection.java
@@ -21,6 +21,10 @@ public interface PriceSelection {
     String getPriceCustomerGroup();
     @Nullable
     String getPriceChannel();
+    @Nullable
+    String getStoreProjection();
+    @Nullable
+    String getLocaleProjection();
 
     static PriceSelectionDsl ofCurrencyCode(final String currencyCode) {
         return PriceSelectionBuilder.ofCurrencyCode(currencyCode).build();
--- a/commercetools-models/src/main/java/io/sphere/sdk/products/search/PriceSelection.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/products/search/PriceSelection.java
@@ -3,6 +3,7 @@ package io.sphere.sdk.products.search;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.money.CurrencyUnit;
+import java.util.List;
 
 /**
  * Parameters to select prices in {@link ProductProjectionSearch}.
@@ -24,7 +25,7 @@ public interface PriceSelection {
     @Nullable
     String getStoreProjection();
     @Nullable
-    String getLocaleProjection();
+    List<String> getLocaleProjection();
 
     static PriceSelectionDsl ofCurrencyCode(final String currencyCode) {
         return PriceSelectionBuilder.ofCurrencyCode(currencyCode).build();
--- a/commercetools-models/src/main/java/io/sphere/sdk/products/search/PriceSelectionBuilder.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/products/search/PriceSelectionBuilder.java
@@ -14,8 +14,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 
-import static java.util.Collections.emptyList;
-
 public final class PriceSelectionBuilder extends Base implements Builder<PriceSelectionDsl> {
     @Nonnull
     private String priceCurrency;
@@ -28,7 +26,7 @@ public final class PriceSelectionBuilder extends Base implements Builder<PriceSe
     @Nullable
     private String storeProjection;
     @Nullable
-    private List <String> localeProjection = emptyList();
+    private List <String> localeProjection = new ArrayList<>();
 
     private PriceSelectionBuilder(@Nonnull final String priceCurrency) {
         this.priceCurrency = priceCurrency;
@@ -97,6 +95,12 @@ public final class PriceSelectionBuilder extends Base implements Builder<PriceSe
 
     public PriceSelectionBuilder localeProjection(@Nullable final List<String> localeProjection) {
         this.localeProjection = localeProjection;
+        return this;
+    }
+
+    public PriceSelectionBuilder localeProjection(@Nullable final String localeProjection) {
+        this.localeProjection = new ArrayList<>();
+        this.localeProjection.add(localeProjection);
         return this;
     }
 
--- a/commercetools-models/src/main/java/io/sphere/sdk/products/search/PriceSelectionBuilder.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/products/search/PriceSelectionBuilder.java
@@ -10,6 +10,7 @@ import io.sphere.sdk.models.Referenceable;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.money.CurrencyUnit;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 
@@ -100,7 +101,7 @@ public final class PriceSelectionBuilder extends Base implements Builder<PriceSe
     }
 
     public PriceSelectionBuilder plusLocaleProjection(@Nullable final String localeProjection) {
-        this.localeProjection = emptyList();
+        this.localeProjection = new ArrayList<>();
         this.localeProjection.add(localeProjection);
         return this;
     }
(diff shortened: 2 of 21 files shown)


commercetools-models/src/main/java/io/sphere/sdk/products/search/PriceSelection.java
```diff
@@ -21,6 +21,10 @@
     String getPriceCustomerGroup();
     @Nullable
     String getPriceChannel();
+    @Nullable
+    String getStoreProjection();
+    @Nullable
+    String getLocaleProjection();
```
katmatt, line 27: The docs explicitly states that we support multiple `localeProjection` query parameters here (see ) and that means that the type should be `List<String>` here instead of just a `String`. This then requires further changes, which I think you fix on your own. I will just add comments below to tell you how the API should look like so that we follow the existing, but undocumented naming conventions of our JVM SDK.

commercetools-models/src/main/java/io/sphere/sdk/products/search/PriceSelectionBuilder.java
```diff
@@ -80,6 +86,16 @@ public PriceSelectionBuilder priceChannel(@Nullable final Referenceable<Channel>
...
 
+    public PriceSelectionBuilder storeProjection(@Nullable final String storeProjection) {
+        this.storeProjection = storeProjection;
+        return this;
+    }
+
+    public PriceSelectionBuilder localeProjection(@Nullable final String localeProjection) {
+        this.localeProjection = localeProjection;
+        return this;
+    }
```
katmatt, line 97: Here we would have to add a `plusLocaleProjection(@Nullable final String localeProjection)` method that adds the given `localeProjection` to the `localeProjection` member variable of this object.

commercetools-models/src/main/java/io/sphere/sdk/products/search/PriceSelectionBuilder.java
```diff
@@ -80,6 +86,16 @@ public PriceSelectionBuilder priceChannel(@Nullable final Referenceable<Channel>
         return this;
     }
 
+    public PriceSelectionBuilder storeProjection(@Nullable final String storeProjection) {
+        this.storeProjection = storeProjection;
+        return this;
+    }
+
+    public PriceSelectionBuilder localeProjection(@Nullable final String localeProjection) {
+        this.localeProjection = localeProjection;
```
katmatt, line 95: Here you would create a new List for the `localeProjection` and add the given `localeProjection` to it.

commercetools-models/src/main/java/io/sphere/sdk/products/search/PriceSelectionQueryParameters.java
```diff
@@ -40,6 +42,8 @@ private PriceSelectionQueryParameters() {
             addParamIfNotNull(resultingParameters, PRICE_COUNTRY, priceSelection.getPriceCountry());
             addParamIfNotNull(resultingParameters, PRICE_CUSTOMER_GROUP, priceSelection.getPriceCustomerGroup());
             addParamIfNotNull(resultingParameters, PRICE_CHANNEL, priceSelection.getPriceChannel());
+            addParamIfNotNull(resultingParameters, STORE_PROJECTION, priceSelection.getStoreProjection());
+            addParamIfNotNull(resultingParameters, LOCALE_PROJECTION, priceSelection.getLocaleProjection());
```
katmatt, line 46: Here we would have to add all items of the `localeProjection` list.

barbara79:
I didn't add the integration test in PriceSelection because we are ignoring all of them, not sure if it's the case to add it for testing the last methods added for the localeProjection and storeProjection

commercetools-models/src/main/java/io/sphere/sdk/products/search/PriceSelectionBuilder.java
```diff
@@ -21,6 +22,10 @@
     private String priceCustomerGroup;
     @Nullable
     private String priceChannel;
+    @Nullable
+    private String storeProjection;
+    @Nullable
+    private List <String> localeProjection;
```
katmatt, line 28: Here we should initialize the list, otherwise calling `plusLocaleProjection` will fail with a `NullPointerException.`:
```java
private List <String> localeProjection = new ArrayList<>();
```

commercetools-models/src/main/java/io/sphere/sdk/products/search/PriceSelectionBuilder.java
katmatt: Here we should add another method :
```java
    public PriceSelectionBuilder localeProjection(@Nullable final String localeProjection) {
        this.localeProjection = new ArrayList<>();
        this.localeProjection.add(localeProjection);
       return this;
   }
```
This would align with our existing code and makes it easier to start with a single value.

commercetools-models/src/main/java/io/sphere/sdk/products/search/PriceSelectionQueryParameters.java
```diff
@@ -35,11 +35,16 @@ private PriceSelectionQueryParameters() {
...
+
         if (priceSelection != null && priceSelection.getPriceCurrency() != null) {
             addParamIfNotNull(resultingParameters, PRICE_CURRENCY, priceSelection.getPriceCurrency());
             addParamIfNotNull(resultingParameters, PRICE_COUNTRY, priceSelection.getPriceCountry());
             addParamIfNotNull(resultingParameters, PRICE_CUSTOMER_GROUP, priceSelection.getPriceCustomerGroup());
             addParamIfNotNull(resultingParameters, PRICE_CHANNEL, priceSelection.getPriceChannel());
+            addParamIfNotNull(resultingParameters, STORE_PROJECTION, priceSelection.getStoreProjection());
+            for (int i = 0; i < priceSelection.getLocaleProjection().size(); i++) {
+                addParamIfNotNull(resultingParameters, LOCALE_PROJECTION, priceSelection.getLocaleProjection().get(i));
+            }
```
jenschude, line 47:
```suggestion
priceSelection.getLocaleProjection.forEach(parameter -> addParamIfNotNull(resultingParameters, LOCALE_PROJECTION, parameter));
```

commercetools-models/src/main/java/io/sphere/sdk/products/search/PriceSelectionBuilder.java
barbara79: ok, I'm gonna update my method `plusLocaleProjection` in this way

commercetools-models/src/main/java/io/sphere/sdk/stores/StoreDraft.java
```diff
@@ -5,21 +5,29 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Locale;
 
 @JsonDeserialize(as = StoreDraftDsl.class)
 @ResourceDraftValue(factoryMethods = {
         @FactoryMethod(parameterNames = {"key"}),
-        @FactoryMethod(parameterNames = {"key", "name"})})
+        @FactoryMethod(parameterNames = {"key", "name", "languages"})})
```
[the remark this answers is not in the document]
barbara79: I'm gonna add a new one with the 2 parameters

commercetools-models/src/main/java/io/sphere/sdk/products/search/PriceSelectionBuilder.java
```diff
@@ -80,6 +89,22 @@ public PriceSelectionBuilder priceChannel(@Nullable final Referenceable<Channel>
...
+        return this;
+    }
+
+    public PriceSelectionBuilder localeProjection(@Nullable final List<String> localeProjection) {
+        this.localeProjection = localeProjection;
+        return this;
+    }
+
+    public PriceSelectionBuilder plusLocaleProjection(@Nullable final String localeProjection) {
+        this.localeProjection = emptyList();
```
katmatt, line 103: Here the code should be:
```java
if (localeProjection == null {
   this.localeProjection = new ArrayList<>();
}
```
And please: never use `emptyList()` if you need a mutable list (as in this case). See also https://www.baeldung.com/java-collections-emptylist-new-list

barbara79: thanks for the explanation,  I'm gonna fix it!

commercetools-models/src/main/java/io/sphere/sdk/products/search/PriceSelectionBuilder.java
jenschude: I just saw the usage also in other builders as they are immutable too.

PR #2086, opened 2020-11-27
Adding RetryableSphereClientWithExponentialBackoff SphereClient utility to avoid duplication in other projects. 

praveenkumarct:
- Adding RetryableSphereClientWithExponentialBackoff, which is a SphereClient utility that provides good defaults for retry logic with considering exponential backoff.

- This code is common code and can be used it everywhere that uses SphereClient. The utility is created in the java-sync, JVM SDK convenience library to avoid code duplication.

 - Updating assertj version for the test cases.

### DIFF
--- a/commercetools-convenience/src/main/java/io/sphere/sdk/client/retry/RetryableSphereClientBuilder.java
+++ b/commercetools-convenience/src/main/java/io/sphere/sdk/client/retry/RetryableSphereClientBuilder.java
@@ -8,16 +8,13 @@ import io.sphere.sdk.client.SphereClientConfig;
 import io.sphere.sdk.http.HttpClient;
 import io.sphere.sdk.models.Base;
 import io.sphere.sdk.retry.RetryAction;
-import io.sphere.sdk.retry.RetryContext;
 import io.sphere.sdk.retry.RetryPredicate;
 import io.sphere.sdk.retry.RetryRule;
 
 import javax.annotation.Nonnull;
-import java.time.Duration;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
-import java.util.function.Function;
 
 import static java.lang.String.format;
 
@@ -25,7 +22,7 @@ import static java.lang.String.format;
  * To create a Sphere Client with retry logic which computes a exponential backoff time delay in milliseconds.
  * And handle all the configurations for the creation of client.
  */
-public final class RetryableSphereClientFactory extends Base {
+public final class RetryableSphereClientBuilder extends Base {
     protected static final long DEFAULT_MAX_DELAY = 60000;
     protected static final long DEFAULT_INITIAL_RETRY_DELAY = 200;
     protected static final int DEFAULT_MAX_RETRY_ATTEMPT = 5;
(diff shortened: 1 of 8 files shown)


jenschude:
There is already the `io.sphere.sdk.client.RetrySphereClientDecorator` available. Which can be configured to do incremental backoff. See https://commercetools.github.io/commercetools-jvm-sdk/apidocs/io/sphere/sdk/client/RetrySphereClientDecorator.html

jenschude:
Hi,

thanks for the PR. But there is already the io.sphere.sdk.client.RetrySphereClientDecorator available. Which can be configured to do incremental backoff. See commercetools.github.io/commercetools-jvm-sdk/apidocs/io/sphere/sdk/client/RetrySphereClientDecorator.html

You can wrap any Client instance using the decorator and apply the Backoff logic using the RetryAction functional interface. Either by using one of the existing methods or by implementing them.

And as long as there is no additional functionality in your PR I would not merge it.

ahmetoz:
Hi @jenschude,

> But there is already the io.sphere.sdk.client.RetrySphereClientDecorator available. Which can be configured to do incremental backoff.

This is not a replacement of the RetrySphereClientDecorator. The code here is providing a utility for the exponential backoff algorithm included using the RetrySphereClientDecorator. 

> You can wrap any Client instance using the decorator and apply the Backoff logic using the RetryAction functional interface. Either by using one of the existing methods or by implementing them.
> And as long as there is no additional functionality in your PR I would not merge it. 

There is, we spend time to make it useful for everyone by adding an exponential backoff algorithm which is not implemented in core SDK, also not easy to know if you have not done it before. I could not see any documentation or suggestion in the core SDK how a good exponential backoff example for the RetrySphereClientDecorator.

The code is not targeting the core SDK, but targeting the convenience package which afaik is for that purpose like before we added useful utilities into this package (convenience) for example correlation id decorator (same you might do by yourself without adding a new code into core package), here the same purpose we are adding a client creation utility that we are sharing in our codebases (more than 5) in professional services to not duplicate the same logic everywhere. 

Note: Before opening this PR, we asked also the @ashishhk if it makes sense.

jenschude:
I would then suggest to split up the class. The calculation of the Backoff could be added to the RetryAction as a new static factory method.

And as the RetryableSphereClientWithExponentialBackoff acts here as a Factory I think the appropiate place would be in the commercetools-convenience module for it.

praveenkumarct:
> I would then suggest to split up the class. The calculation of the Backoff could be added to the RetryAction as a new static factory method.
> 
> https://github.com/commercetools/commercetools-jvm-sdk/blob/master/commercetools-sdk-base/src/main/java/io/sphere/sdk/retry/RetryAction.java
> 
> And as the RetryableSphereClientWithExponentialBackoff acts here as a Factory I think the appropiate place would be in the commercetools-convenience module for it.
> 
> https://github.com/commercetools/commercetools-jvm-sdk/tree/master/commercetools-convenience/src/main/java/io/sphere/sdk/client

praveenkumarct:
@jenschude, Changes are committed as we discussed. Could you please review this PR.

jenschude (approved):
I renamed the factory class. Added a factory method to the RetryAction to create an exponential backoff RetryAction directly and adjusted the tests

Thanks for the contribution 👍

PR #2106, opened 2021-01-19
support ResourceIdentifier to set customer groups

### LINKED ISSUE #2072: SetCustomerGroup action needs to support ResourceIdentifier.
Issue https://jira.commercetools.com/browse/SUPPORT-10261

Update action SetCustomerGroup is not easy to create because it needs a Referenceable object in its constructor.

http://commercetools.github.io/commercetools-jvm-sdk/apidocs/io/sphere/sdk/customers/commands/updateactions/SetCustomerGroup.html

So if you don't have a customer group to update it is not easy to create a Referenceable object as it has not a constructor,

So library user needs to do a hacky way of implementing an anoynmous class of ResourceImpl like below:

```
return new ResourceImpl<CustomerGroup>(null, null, null, null) {
@Override
public Reference<CustomerGroup> toReference()

{ return Reference.of(CustomerGroup.referenceTypeId(), resourceIdentifier.getId()); }
};
```

So in this case it's better to change the action constructor with
`ResourceIdentifier<CustomerGroup>`, similar to other update actions in the library.

jenschude:
Closes #2072

# Checklist

- [x] Update release Notes

### DIFF
--- a/commercetools-models/src/main/java/io/sphere/sdk/customers/commands/updateactions/SetCustomerGroup.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/customers/commands/updateactions/SetCustomerGroup.java
@@ -4,6 +4,7 @@ import io.sphere.sdk.commands.UpdateActionImpl;
 import io.sphere.sdk.customergroups.CustomerGroup;
 import io.sphere.sdk.customers.Customer;
 import io.sphere.sdk.models.Referenceable;
+import io.sphere.sdk.models.ResourceIdentifiable;
 import io.sphere.sdk.models.ResourceIdentifier;
 
 import javax.annotation.Nullable;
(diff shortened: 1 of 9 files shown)


commercetools-models/src/main/java/io/sphere/sdk/customers/commands/updateactions/SetCustomerGroup.java
```diff
@@ -22,19 +23,20 @@
...
         super("setCustomerGroup");
-        this.customerGroup = Optional.ofNullable(customerGroup)
-                .map(x -> x.toReference().filled(null))
-                .orElse(null);
+        this.customerGroup = customerGroup;
     }
 
-    public static SetCustomerGroup of(@Nullable final Referenceable<CustomerGroup> customerGroup) {
-        return new SetCustomerGroup(customerGroup);
+    public static SetCustomerGroup of(@Nullable final ResourceIdentifiable<CustomerGroup> customerGroup) {
```
barbara79, line 31: I found the same method and action in the cart, for the same purpose of the ticket could make sense to do it for the SetCustomerGroup in the Cart action @ashishhk ? It would be difficult to pass a Referenceable in the of method, isn't it?
this is the part of the code 
    `public static SetCustomerGroup of(@Nullable final Referenceable<CustomerGroup> customerGroup) {
        final ResourceIdentifier<CustomerGroup> resourceIdentifier = Optional.ofNullable(customerGroup)
                .map(ResourceIdentifiable::toResourceIdentifier)
                .orElse(null);
        return new SetCustomerGroup(resourceIdentifier);
    }
`

jenschude: a Referenceable extends ResourceIdentifiable. So totally valid

barbara79 (approved):

PR #2133, opened 2021-04-16
fixed Tiers field not included in PriceDraft

barbara79:
# Description
fix bug tiers not included in the PriceDraftBuilder and as withTiers method in PriceDraft

Closes #

# Checklist

- [x] Update release Notes

### DIFF
--- a/commercetools-models/src/main/java/io/sphere/sdk/products/PriceDraftDsl.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/products/PriceDraftDsl.java
@@ -148,6 +148,10 @@ public final class PriceDraftDsl extends Base implements PriceDraft {
         return PriceDraftBuilder.of(this).discounted(discounted).build();
     }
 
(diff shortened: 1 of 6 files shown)


commercetools-models/src/main/java/io/sphere/sdk/products/PriceDraftDsl.java
```diff
@@ -148,6 +148,10 @@ public PriceDraftDsl withDiscounted(@Nullable final DiscountedPrice discounted)
         return PriceDraftBuilder.of(this).discounted(discounted).build();
     }
 
+    public PriceDraftDsl withTiers(@Nullable final List<PriceTier> tiers) {
+        return PriceDraftBuilder.of(this).tiers(tiers).build();
+    }
+
```
barbara79, line 154: I found out the this method was missing too, I think is correct to implement it, what do you think about?
I also added the test to test if it's working properly

jenschude (approved):

PR #2165, opened 2021-08-03
Store shoppinglist

barbara79:
# Description

Closes #

# Checklist

- [x] Update release Notes

### DIFF
--- a/commercetools-models/src/main/java/io/sphere/sdk/shoppinglists/ShoppingList.java
+++ b/commercetools-models/src/main/java/io/sphere/sdk/shoppinglists/ShoppingList.java
@@ -74,7 +74,9 @@ public interface ShoppingList extends Resource<ShoppingList>, Custom, WithKey {
     @Nullable
     String getAnonymousId();
 
+    @HasUpdateAction
     @Nullable
+    @IgnoreInQueryModel
     KeyReference<Store> getStore();
 
     /**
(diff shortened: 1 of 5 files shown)


commercetools-models/src/main/java/io/sphere/sdk/shoppinglists/ShoppingList.java
```diff
@@ -76,6 +74,11 @@
     @Nullable
     String getAnonymousId();
 
+    @HasUpdateAction
+    @Nullable
+    @IgnoreInQueryModel
```
jenschude, line 79: why did you flagged it to be ignored in the query model?

commercetools-models/src/test/java/io/sphere/sdk/shoppinglists/commands/ShoppingListUpdateCommandIntegrationTest.java
```diff
@@ -3,6 +3,7 @@
 import io.sphere.sdk.models.LocalizedString;
 import io.sphere.sdk.shoppinglists.*;
 import io.sphere.sdk.shoppinglists.commands.updateactions.*;
+import io.sphere.sdk.stores.StoreFixtures;
```
jenschude, line 6: seems you missed a test with a store ;)

jenschude (changes requested):
Seems the whole actions to set and query the shopping list based on the store are missing

https://docs.commercetools.com/api/releases/2021-04-23-add-store-to-shopping-list

commercetools-models/src/main/java/io/sphere/sdk/shoppinglists/ShoppingList.java
barbara79: because without this annotation it gives me the error "java: cannot find symbol" in the querys, querybuilders, queryImpls, draftbuilders and so on classes

commercetools-models/src/test/java/io/sphere/sdk/shoppinglists/commands/ShoppingListUpdateCommandIntegrationTest.java
barbara79: I just saw that if the method has been automatically generated with the annotation @HasUpdateAction there are no tests related like if you see the getAnonymousId() ;)

